PR #40, opened 2015-05-28
test-msupdates

dilorenzo1987:
Disable alternative method as its not working correctly for uninstalled
patches


mmashwani:
Rejecting this pull request as I would rather fix the bug then just delete the alternate update search method.

PR #166, opened 2016-04-18
Fixing issue Error Test-MSUpdates (Refer to : #164)

Reclad:
For fixing issue : https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/164

I added to -match parenthesis between KBNumber because it always showed with parenthesis in title of KBUpdate. I don't know if it same in all countries (but i think yes).


Reclad:
Fixed in 3.6.9 : https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/master/CHANGELOG.txt#L6

Issue #255, opened 2017-03-15
-CheckDiskSpace sets the value in $closeAppsCountdown for some reason...


Latupho:
For some reason someone have set the "-RequiredDiskSpace" on the "-CheckDiskSpace" property, like so `-CheckDiskSpace '10240' `. That gave a odd error:
"The close applications countdown time cannot be longer than the timeout specified in the XML configuration for installation UI dialogs to timeout."

**Code:**
`Show-InstallationWelcome -CloseApps 'VS_Control=Vico Schedule Planner,AcroRd32' `
-CheckDiskSpace '10240' `
-PersistPrompt `
-PromptToSave `
-DeferDays 7 `
-AllowDeferCloseApps `
-MinimizeWindows $false
`

I traced it down, and do to the Parameter "-CheckDiskSpace" is set to be a `[switch]$CheckDiskSpace ` it is not making any checks on the value input is `[boolean]`.

The value is being passed to the variable ` $closeAppsCountdown` for some odd reason, and this triggers the error described above.

I will create a EX on a fix to this a little lather.

Latupho:
I have added a Pull Request on this with a ex. on a fix.
https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/pull/256

seanlillis:
This is incorrect usage of the parameters. The parameter types and usage have been documented:

.PARAMETER CheckDiskSpace **<switch>**
	Specify whether to check if there is enough disk space for the installation to proceed.
	If this parameter is specified without the RequiredDiskSpace parameter, the required disk space is calculated automatically based on the size of the script source and associated files.
.PARAMETER RequiredDiskSpace **<integer>**
	Specify required disk space in MB, used in combination with CheckDiskSpace.

seanlillis:
I have pushed a fix for this to use a Parameter set that requires both parameters to be used together and also performs validation on the CheckDiskSpace parameter to ensure it is of type [switch].

PR #256, opened 2017-03-15
More info on issue #255 - If you set a value (Not Bool on -CheckDiskSpace) you get an error.

Latupho:
As I think I'm not the only one having some setting this parmeter  in a wrong way, I have added a check on it.


Latupho:
I wuld like to do something like:
`[Parameter(Mandatory=$false)]
[ValidateScript({ $_ -eq [boolean] }) ]
[switch]$CheckDiskSpace = $false,`

But can't figure it out, if you can at all...

seanlillis:
I have pushed a fix for this to use a Parameter set that requires both parameters to be used together and also performs validation on the CheckDiskSpace parameter to ensure it is of type [switch].

PR #259, opened 2017-03-27
Added countdown timer to Show-InstallationPrompt

bruckect:
I added a ShowCountdown switch to Show-InstallationPrompt that counts down like Show-InstallationWelcome does using the value that is supplied from $timeout. I hope others would find this useful and hope it can be included in a future release of the toolkit.


sintaxasn:
This issue was moved to https://discourse.psappdeploytoolkit.com/t/added-countdown-timer-to-show-installationprompt/1217

PR #285, opened 2017-06-23
Modified Show-DialogBox TopMost Switch

NinjaKittens14:
The Show-DialogBox TopMost switch does not work as it is set to a default value of $true, making it always on top even without the switch. I set the default value to $false so that the dialog box will always be on top only if the TopMost switch has been used. Otherwise, the window will be based on focus when the switch is unused.


seanlillis:
Hi, thanks for flagging this. Your fix is valid, however we prefer to keep the same default behaviour and stay consistent with the other UI functions in the toolkit, so I'm changing the TopMost variable to a Boolean in this function, which will allow you to toggle it off and on as necessary.

PR #316, opened 2017-11-30
Add 'Unattended' switch for DeployMode

evgenyfedorov2:
Added new switch value for DeployMode parameter - "Unattended".
Now, if you have Show-InstallationWelcome and Show-InstallationProgress function calls in the script, it will work like this:
1. "Silent", "Non-Interactive" - nothing is visible
2. "Interactive" - both Show-InstallationWelcome and Show-InstallationProgress pop-up
3. "Unattended" - Show-InstallationWelcome is invisible (run in Silent mode), Show-InstallationProgress is visible


sintaxasn:
This issue was moved to https://discourse.psappdeploytoolkit.com/t/add-unattended-switch-for-deploymode/1169

Issue #335, opened 2018-02-24
Deploy-Application.exe Does Not Support Spaces In Parameters


KingLinkTiger:
This issue is somewhat relevant to Issue #200, adding custom parameters to Deploy-Application.ps1. If you add these custom parameters Deploy-Application.exe does not support spaces in them, and if you invoke Deploy-Application.exe with these parameters your script will fail to run. Another thing to now is since the failure occurs before Deploy-Application.ps1 is invoked no logging will occur and the user will not know why their script is failing to run.

---

How to reproduce the issue:

For the following examples I'm running the following command:
```
.\Deploy-Application.exe -FileName "Hello World" -FilePath "C:\Program Files\Hello World"
```

Currently how Deploy-Application.exe handles parameters. This is the output of Deploy-Application.exe that runs Deploy-Application.ps1.
```
C:\Windows\System32\WindowsPowerShell\v1.0\PowerShell.exe -ExecutionPolicy Bypass -NoProfile -NoLogo -WindowStyle Hidden -Command & { & 'C:\Users\KLT\Downloads\PSAppDeployToolkit-master\PSAppDeployToolkit-master\Sources\Deploy-Application\Deploy-Application\bin\Debug\Deploy-Application.ps1' -FileName Hello World -FilePath C:\Program Files\Hello World; Exit $LastExitCode }
```

Ideally how Deploy-Application.exe should handle parameters:

```
C:\Windows\System32\WindowsPowerShell\v1.0\PowerShell.exe -ExecutionPolicy Bypass -NoProfile -NoLogo -WindowStyle Hidden -Command & { & 'C:\Users\KLT\Downloads\PSAppDeployToolkit-master\PSAppDeployToolkit-master\Sources\Deploy-Application\Deploy-Application\bin\Debug\Deploy-Application.ps1' -FileName "Hello World" -FilePath "C:\Program Files\Hello World"; Exit $LastExitCode }
```
Notice how currently the quotation marks are removed. I believe the issues to stems from line 26 of [DeployApplication.cs](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/master/Sources/Deploy-Application/Deploy-Application/DeployApplication.cs) where Environment.GetCommandLineArgs() is used. The [.Net documentation](https://msdn.microsoft.com/en-us/library/system.environment.getcommandlineargs(v=vs.110).aspx) shows in its examples that this strips quotation marks when creating the string array.

---

[...]

Issue #200, opened 2016-08-12
Any way to pass arbitrary parameters to Deploy-Application.exe?


DSBloom:
Is there any way to pass "custom" parameters to Deploy-Application.exe? 

For example:
`Deploy-Application.exe /MyParam`

So that later in the Deploy-Application.ps1 script we can perform certain actions if /MyParam was used?

Whenever I try and do something like that Deploy-Applicaiton.exe just doesn't launch.

That-Annoying-Guy:
I have never used DeployApplication.exe and I'm weak in CS but from what I can tell from the source code in DeployApplication.cs, it seem you need to do it this way:

Deploy-Application.exe -File "<Path to Deploy-Application.ps1>" -Command "Params"

But this is but a pure guess.  Deploy-Application.exe is not a fully supported feature of PSADT. Your best bet is to poke around the website for other mentions of Deploy-Application.exe.

nadcraker:
Thanks for the tip. I'll look into that. But it looks like the Deploy-Application.ps1 file validates the parameters, and won't accept anything that it is not expecting.
`

	Param (
		[Parameter(Mandatory=$false)]
		[ValidateNotNullorEmpty()]
		[string]$StatusMessage = $configProgressMessageInstall,
		[Parameter(Mandatory=$false)]
		[ValidateSet('Default','BottomRight')]
		[string]$WindowLocation = 'Default',
		[Parameter(Mandatory=$false)]
		[ValidateNotNullorEmpty()]
		[boolean]$TopMost = $true
	)`

That-Annoying-Guy:
So.
Just follow the pattern and add the parameters you need to Deploy-Application.ps1
I do this all the time.

I don't think this is a bug in PSADT other than Deploy-Application.exe needs to be documented. I know there was talk to officially support it in a future major version like v4. 
Time will tell.

nadcraker:
Well, I was hoping to leave PSADT as 'stock' as possible, but your suggestion is simple enough to implement and see how it goes.

So I'll change my Deploy-Application.ps1 CMDLETBINDING section to look like this:


'Param (
	[Parameter(Mandatory=$false)]
	[ValidateSet('Install','Uninstall')]
	[string]$DeploymentType = 'Install',
	[Parameter(Mandatory=$false)]
	[ValidateSet('Interactive','Silent','NonInteractive')]
	[string]$DeployMode = 'Interactive',
[...]

PR #336, opened 2018-02-24
Proposed Fix for Issue #335

KingLinkTiger:
Proposed Fix for Issue #335 and in extension #200.

Change the use of Environment.GetCommandLineArgs to Environment.CommandLine and then manually parse it in the script. We'll need to split the string on the hyphen character as this is the character used to denote arguments in PowerShell. This will cause an issue with strings that contain hyphens such as "Deploy-Application", but if we split on a space and a hyphens (ex " -") as all arguments have to be separated by spaces preceding the hyphens. Line 84/85 has to be modified as the split command I'm using removes the hyphen so I have to put it back in somewhere.

A better solution would be to use Regex to do the splitting to preserve the hyphen but that requires an additional reference not currently being used. I believe this subject should be discussed further for a more ideal solution.


sintaxasn:
This issue was moved to https://discourse.psappdeploytoolkit.com/t/proposed-fix-for-issue-335/1161

PR #371, opened 2018-08-18
[Get-InstalledApplication] Search by Publisher

khaffner:
Allows to search by publisher name. Useful in some edge cases like "Please uninstall everything from Amazon". 

I personally think Get-InstalledApplication and Remove-MSIApplications need an overhaul. Get-InstalledApplication should just get everything, then use the pipe for filtering before maybe giving it to Remove-MSIApplications for proper uninstallation. But changing as much as I imagine would break a lot in the toolkit. Sadly Get-Package and Uninstall-Package do not give the same properties as the cmdlets in PSADT.


sintaxasn:
This issue was moved to https://discourse.psappdeploytoolkit.com/t/get-installedapplication-search-by-publisher/1138

PR #414, opened 2019-06-12
Get-InstalledApplication should only replace invalid characters

CatalinSoare:
Only replace invalid characters, not an entire segment since that would have removed localized app display names.
Removing the replacement entirely is not a good idea since we may have control characters.


sintaxasn:
Nice work, thank you :)

PR #419, opened 2019-07-13
Show-InstallationProgress fix

luki1412:
The dynamic banner size broke this function, making it not display anything and just post a long message into the log. This is a fix for it.


sintaxasn:
Great work, thank you!

Issue #425, opened 2019-08-21
Application deployment executable error


kevball2:
Running into an issue with the latest zip download of the toolkit. Issue is with launching the application executable but getting an error "There is no Unicode byte order mark. Cannot Switch to Unicode"

![image](https://user-images.githubusercontent.com/22545150/63462471-a3f12c00-c420-11e9-9939-398a9d93b21d.png)

Tried a couple downloads but no luck getting even the default package to run

luki1412:
Thats because of the config xml file. When you download it from here, its encoded in UTF8 however in the file UTF16 is specified so you get this error. Regular releases should contain file with correct encoding. Dont know why UTF16 was necessary honestly. I can create a pull request.

luki1412:
This was fixed.

PR #426, opened 2019-09-05
Change encoding to fix an issue

luki1412:
There is an issue with this file because you download everything here encoded as utf8 and this file specifies that its utf16. The encoding doesnt match and so the toolkit throws an error.

Issue #425


sintaxasn:
Thanks! I was hoping to get to this tomorrow :)

Issue #436, opened 2019-10-02
AppDeployToolkitMain.ps1 failed to load:  You must provide a value expression on the right-hand side of the '-' operator


z442a9:
For some reason some machines are giving me this error and I'm not sure what is the issue. 
AppDeployToolkit\AppDeployToolkitMain.ps1] failed to load:  You must provide a value expression on the right-hand side of the '-' operator.
 
At \AppDeployToolkit\AppDeployToolkitMain.ps1:7520 char:17

+             If ($Action - <<<< in 'PintoStartMenu', 'UnpinfromStartMenu')
At line:1 char:25
+ .\deploy-application.ps1 <<<< 
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorExcep 
   tion
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorExceptio 
   n,Deploy-Application.ps1
 
Any ideas? Some machines run this same script fine.

snowmountainer:
I noticed the same error on a computer running Windows 7 64-bit. 
-------------------------------------------------------------------------
$PSVersionTable

Name                           Value                                                                                                                                                                                                       
----                           -----                                                                                                                                                                                                       
CLRVersion                     2.0.50727.8806                                                                                                                                                                                              
BuildVersion                   6.1.7601.23403                                                                                                                                                                                              
PSVersion                      2.0                                                                                                                                                                                                         
WSManStackVersion              2.0                                                                                                                                                                                                         
PSCompatibleVersions           {1.0, 2.0}                                                                                                                                                                                                  
SerializationVersion           1.1.0.1                                                                                                                                                                                                     
PSRemotingProtocolVersion      2.1

SierraEcoBravo:
Someone please correct me but PowerShell 2 doesn't support the -in Comparison operator. Check the version of PowerShell on the systems with the error.

Good luck.

luki1412:
> Someone please correct me but PowerShell 2 doesn't support the -in Comparison operator. Check the version of PowerShell on the systems with the error.
> 
> Good luck.

That is correct. The -in parameter isnt supported under the ps 2.0 version that comes with win 7. There is a simple fix for this case but Im not sure if there are any other changes we would need to do to support this version.

snowmountainer:
[...]

PR #437, opened 2019-10-07
Removed the -in operator

luki1412:
Powershell 2.0 doesn't support the **-in** parameter(at least not with the build that gets released with win 7).

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -7517,7 +7517,7 @@ Function Set-PinnedApplication {
 				Throw "Action [$Action] not supported. Supported actions are [$($Verbs.Keys -join ', ')]."
 			}
 
-			If ($Action -in 'PintoStartMenu', 'UnpinfromStartMenu')
+			If ($Action.Contains("StartMenu"))
 			{
 				If ([int]$envOSVersionMajor -ge 10)	{
 					If ((Get-Item -Path $FilePath).Extension -ne '.lnk') {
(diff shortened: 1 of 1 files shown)


RobinBeismann:
This should fix #436

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -7535,7 +7535,7 @@ Function Set-PinnedApplication {
 
 				Invoke-Verb -FilePath $FilePath -Verb $PinVerbAction
 			}
-			ElseIf ($Action -in 'PintoTaskbar', 'UnpinfromTaskbar') {
```
RobinBeismann, line 7538: Please add a comment on why we're not using -in but instead .Contains() so the devs later on know it.

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -7517,7 +7517,7 @@ Function Set-PinnedApplication {
 				Throw "Action [$Action] not supported. Supported actions are [$($Verbs.Keys -join ', ')]."
 			}
```
RobinBeismann, line 7519: Please add a comment on why we're not using -in but instead .Contains() so the devs later on know it.

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -7517,7 +7517,7 @@ Function Set-PinnedApplication {
...
 
-			If ($Action -in 'PintoStartMenu', 'UnpinfromStartMenu')
+			If ($Action.Contains("StartMenu"))
```
luki1412, line 7520: Replacing the -in parameter because it is not supported under Powershell 2.0

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -7535,7 +7535,7 @@ Function Set-PinnedApplication {
 
 				Invoke-Verb -FilePath $FilePath -Verb $PinVerbAction
 			}
-			ElseIf ($Action -in 'PintoTaskbar', 'UnpinfromTaskbar') {
+			ElseIf ($Action.Contains("Taskbar")) {
```
luki1412, line 7538: Replacing the -in parameter because it is not supported under Powershell 2.0

Issue #452, opened 2019-11-13
Execute-MSI does not replace bad characters in log names properly and Zip-compressions of logfolder might fail


Basti890:
I have Discovered this when using `Execute-MSI uninstall "PRODUCTCODE"` the logname is generatet by using the data from the registry entry of this software.
By using this characters like : ®™ are not removed from the Log-Name and the when using Zip-compression for the logs the operation fails due bad filenames.
So for this Software I get an error after the uninstall.
 
![Window-2019-11-13 10_07_34](https://user-images.githubusercontent.com/11937981/68749369-64407780-05fe-11ea-97de-fe0b1531cf5d.png)

Basti890:
As a quick fix I'm now using the -LogName Parameter to set for this specif uninstall operation a custom LogName.

luki1412:
Fixed.

PR #462, opened 2019-12-16
Add function Remove-InvalidFileNameChars

luki1412:
This patch adds the function Remove-InvalidFileNameChars to the toolkit and is used in places where filename was being validated before. I also added removal of invalid chars to the New-Zipfile command since it doesn't have one, causing it to fail while creating zip files when an invalid character is specified.

#452 
possibly #397

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -1,4 +1,4 @@
-﻿<#
+<#
 .SYNOPSIS
 	This script contains the functions and logic engine for the Deploy-Application.ps1 script.
 	# LICENSE #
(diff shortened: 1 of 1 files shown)


cyborgrd:
I get an error with the new Remove-InvalidFileNameChars 
PArameter 'replace' not found:

[Uninstallation] :: Error Record:
-------------

Message        : Es wurde kein Parameter gefunden, der dem Parameternamen "replace" entspricht.
InnerException : 

FullyQualifiedErrorId : NamedParameterNotFound,Remove-InvalidFileNameChars
ScriptStackTrace      : bei Execute-MSI<Process>, 
                        \AppDeployToolkitMain.ps1: Zeile 2348

cyborgrd:
indeed there is no parameter replace in the function Remove-InvalidFileNameChars

luki1412:
> indeed there is no parameter replace in the function Remove-InvalidFileNameChars

Yes, there isn't. It's an operator following the function. We just need a set of brackets there. I'll create a PR for it.

Issue #125, opened 2015-12-04
Execute-Process exits script on file not found for target executable


amotaz:
Execute-Process exits script on file not found for target executable.
This is regardless of ContinueOnError parameter. We can consider the implications of adding a check for ContinueOnError before exiting the script. 

Marking as enhancement since the definition of ContinueOnError currently is strictly defined to a reaction to the executable exit code.

amotaz:
We may modify the continueonerror behavior long term, but not in the next minor release

sintaxasn:
This issue was moved to https://discourse.psappdeploytoolkit.com/t/execute-process-exits-script-on-file-not-found-for-target-executable/1194

luki1412:
#479 modifies behavior of -ContinueOnError and another PR adds ability to not exit script on error.

PR #479, opened 2020-04-03
BREAKING CHANGE: Modify Execute-Process -ContinueOnError behavior

luki1412:
Currently Execute-Process is one of the functions that fails on error even if ContinueOnError is specified because this parameter is used for ignoring exit codes and not handling errors. This PR fixes it by adding the ability to ignore all exit codes through -IgnoreExitCodes "*" while -ContinueOnError will be used for ignoring errors that occurred inside the function as expected.

Also fixes a couple of typos and fixes -PassThru because currently, it makes the function ignore exit codes. 

Rewords log entry for when this function is used for an MSI installation but another installation is already running.

#125


ocsscott:
does this have to be compiled into the main exe or do we just need to replace the 1 ps1 file?

thanks

________________________________

Issue #398, opened 2019-03-04
executeasuser.vbs naming


tda90:
Hello Everyone,
I've a little problem with the generated "PSAPPDEPLOYTOOLKIT-EXECUTEASUSER.vbs" file.
Sine we're implementing AppBlocker I will have to sign or to whitelist trough hash.

Now we have a couple of applications which need the executeasuser.vbs.
I've seen that the name of this VBS is always the same (and it's hardcoded in the AppDeployToolkitMain.ps1).

Now my question is... what happen if you have 2 application which need executeasuser.vbs?
Since the executeasuser.vbs has always the same name and is always in the same location? (C:\Users\PUBLIC\PSAPPDEPLOYTOOLKIT\PSAPPDEPLOYTOOLKIT-EXECUTEASUSER.VBS) ?

Thank you for helping.
Regards

luki1412:
Execute-ProcessAsUser is used for executing an application in another user's context. That is why it creates a new vbs file on every execution. However it keeps overwriting the same file. After task scheduler runs the process and gets an exit code, the task gets deleted. The file stays there. It should be cleaned up.

seanlillis:
Closing as an answered question.

Issue #482, opened 2020-04-13
Potential security risk by using C:\Users\Public\PsAppDeployToolkit


renini:
The current defaults of having Toolkit_TempPath pointing to C:\Users\Public might introduce a security risk.

RobinBeismann:
Can you elaborate why? C:\Users\Public is the library for Documents/Shortcuts that are available to everyone.
However NTFS Permissions are by default restricted to the system/administrators plus the creator of the files.

cmcit-bferg:
The CVE you referenced isn't even published yet?

renini:
> Can you elaborate why? C:\Users\Public is the library for Documents/Shortcuts that are available to everyone.
> However NTFS Permissions are by default restricted to the system/administrators plus the creator of the files.

Because most of the times the files are executed as LocalSystem, these are better placed somewhere in C:\Windows or C:\Program Files\. 

By changing this, this will also allow for the default AppLocker policy and probably also closes #398 and #447.

RobinBeismann:
> The CVE you referenced isn't even published yet?

@renini, are you sure this is the correct CVE? I can't find CVE-2020-10962 on the internet either.

renini:
Yes that is correct, it’s reserved to track this specific issue.

luki1412:
This CVE is a reserved one - https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-10962
So not submitted one.

Also, this security risk only affects toolkits that use the function **Show-InstallationWelcome** with the parameter **-BlockExecution** or the internal function **Block-AppExecution** AND are running with **elevated permissions** AND can be exploited only **while the script is running**.

_The function should be rewritten in the version 4.0 but for right now you should change the path in affected packages/applications where you use the script under these conditions by changing the temp path inside the xml config to a folder not accessible by regular users._

Issue #447, opened 2019-10-29
ActiveSetup not triggering for installing user due to AppLocker


That-Annoying-Guy:
When you install a PSADT script and it uses Set-ActiveSetup, it will run what is placed in the StubPath value for the user that is running the PSADT script. This saves the installing user a reboot.

This works fine until your organization decides to use Windows 10's AppLocker feature.
AppLocker is designed to stop scripts and EXE from running from the User's profile.
AppLocker is great to block scripts that the user downloads (knowingly or not) and user-created scripts. Usually AppLocker is configured to whitelist locations where users have no permissions.  

FYI: When AppLocker blocks a script, it lies to the process that launch the script and returns exitcode 0. IOW: The script never runs but the parent process thinks it did.

For PSADT to run the StubPath it creates a **c:\users\PSADT\$appDeployToolkitName-ExecuteAsUser.vbs** script. Because the VBS script is created in a profile folder, AppLocker blocks the script and returns exitcode 0 to the main PSADT script.

Since we configured AppLocker to allow script to run in C:\ProgramFiles\ for other applications,
OUR solution was to move the PSADT folder to C:\ProgramFiles\ 
This works because PSADT is always run elevated (at least in my shop) and therefore has the permissions to write to C:\ProgramFiles\.

The **c:\users\PSADT** path is set in AppDeployToolkitConfig.xml so we changed:

		<Toolkit_TempPath>$envSystemDrive\Users\Public</Toolkit_TempPath>
		
to:

		<Toolkit_TempPath>$envProgramFiles</Toolkit_TempPath>

This fixed the issue. 

It's up to the main authors of PSAppDeployToolkit to decide if this should be the new default or not.

I hope other users of AppLocker can chime in to validate or put down my solution.

luki1412:
AppLocker ignores scripts running inside windows folder and program files by default. I would rather use Windows folder. Also the name for the vbs script is way too long IMO.

luki1412:
The default temp path was changed to Windows\Temp when running with admin rights. Which is not accessible to user at all. So the script will run fine but the ActiveSetup script/app has to be elsewhere, accessible to user but without write access.

PR #483, opened 2020-04-13
Fix insecure defaults that could allow elevation of privilege

renini:
Fix for issue #482 insecure defaults that could allow elevation of privilege (CVE-2020-10962)


renini:
With this PR we can probably also close #398 and #447.

Issue #503, opened 2020-05-17
Question on "Write-Log"-Update regarding "Source"-Parameter


NordicAT:
Hello,

I just tested the latest 3.8.2 PSADTK-Version coming from 3.8.0.

I make extensively usage of "**Write-Log**" in all my packages and there seems to be a change in the function:

Command:
`Write-Log -Message "$appVendor $appName (Version: $appVersion) is beeing installed ..." -Severity 1`

Output-Before:
> [05-17-2020 12:20:49.124] [Installation] :: VideoLAN VLCPlayer (Version: 3.0.10) is beeing installed ...

Output-After:
> [05-17-2020 12:19:25.723] [Installation] [Unknown] :: VideoLAN VLC Player (Version: 3.0.10) is beeing installed ...

---

Of course this is only a cosmetical issue, but now the logs are full of **[Unknown]** -strings.

The changelog of 3.8.1 says:
> Added default -Source parameter value for Write-Log to fix CMTrace not displaying severity

The word-manual says for the source-parameter:

> -Source <String>
> The source of the message being logged.

---

I am sorry, but I am still not sure, what the parameter "Source" is meant for?

What should I use for the source to get rid of [unknown]? (Recommendation)
Ideally the log-line should look like before.

Thank you

luki1412:
This parameter is used for the Source of the log entry. What is the origin of this log entry.
If the Source is missing, CMTrace doesn't display the entry as red or yellow for errors and warnings.
So if you omit this parameter, a value of Unknown is used so its not empty and the log is displayed correctly. You usually want to put the name of the function there.

redcrossevvy:
Would it be possible, perhaps as a feature request, to default this to $appName, to reduce redundant typing?


On May 17, 2020, at 8:45 AM, Ultimate Luki <notifications@github.com> wrote:

﻿

This parameter is used for the Source of the log entry. What is the origin of this log entry.
If the Source is missing, CMTrace doesn't display the entry as red or yellow for errors and warnings.
So if you omit this parameter, a value of Unknown is used so its not empty and the log is displayed correctly. You usually want to put the name of the function there.

—
You are receiving this because you are subscribed to this thread.
[...]

PR #505, opened 2020-05-18
Add config paths without admin rights

luki1412:
Adds config paths for installations where RequireAdmin is False.

Also changes default -Source value for Write-Log to a more useful value: it will be set to Parent's function name and if not inside a function then the name of the script. If not inside a function or a script, then the value will be set to 'Unknown'

#482 
#503


redcrossevvy:
Super – thank you!

PR #506, opened 2020-05-18
Fix bug with $envOSVersionRevison and Normalized ProgramFiles and CommonProgramFiles Path Variables

mmashwani:
- Fix issue with $envOSVersionRevision not being populated on Windows 8 because the "UBR" registry value is only being checked if "$envOSVersionMajor = 10"

- Fixed an issue where the correct ProgramFiles and CommonProgramFiles paths were not populated if the Toolkit was launched from a 32-bit process on a 64-bit OS.


sintaxasn:
A very pleasant surprise to see this Mo! Thanks very much and let me know if you want to discuss further :)

mmashwani:
Hey Dan,

Of course, no hard feelings man. I found the bug in the course of some dev
work and I wouldn’t want anyone else tripping up over it. I will always
provide fixes for any issues I find.

I’m happy to see lots of activity with the project lately. Seems like you
guys are doing some great work!

Regards,
Mo

Issue #513, opened 2020-05-27
Return Process Id from Execute-Process when -PassThru is specified


iainbrighton:
**Describe the solution you'd like**
When launching `Execute-Process` with `-NoWait -PassThru` parameters, it would be beneficial to have visibility of the process that is launched so that it can be monitored or waited upon later during the install script. Adding the process Id (PID) to the $ExecutionResults object returned would enable this.

**Additional context**
The Citrix Receiver/Workspace App requires the mouse to be moved during the USB component installation. Having the ability to start the install process and programmatically move the mouse cursor until the process has exited would be possible with this addition.

I'm happy to submit a PR for this if the general consensus is that this is a worthwhile addition.

Thanks,
Iain

luki1412:
I agree. Some people might decide to check on process' progress after it has been started with -NoWait

PR #515, opened 2020-05-29
Add a return value if NoWait and PassThru are specified

luki1412:
Allows Execute-Process to return an object if **PassThru and NoWait are both specified**. Normally we don't have an output since the process hasn't stopped yet but instead of returning nothing we can return process id, process name and process handle if a user wants to check something through them.

#513

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -2883,7 +2883,7 @@ Function Execute-Process {
 .PARAMETER NoWait
 	Immediately continue after executing the process.
 .PARAMETER PassThru
-	Returns ExitCode, STDOut, and STDErr output from the process.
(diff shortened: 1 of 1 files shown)


redcrossevvy:
Goodness... that's a stretch. Now I will have to write code to determine if I'm looking at an error code or a process ID.  That seems messy to me.  Since the function returns an object can you implement additional members for the object so it is clearer what we are looking at?

luki1412:
> Goodness... that's a stretch. Now I will have to write code to determine if I'm looking at an error code or a process ID. That seems messy to me. Since the function returns an object can you implement additional members for the object so it is clearer what we are looking at?

What? No you dont have to write anything, you either use NoWait or you dont. If you do, you would have gotten nothing prior to this PR. Now you do but only if its NoWait, other wise the functionality doesn't change. Why would you even need to check the output object? You are not sure whether you used -NoWait ?

PR #525, opened 2020-06-26
Rename Restart Later to Minimize

### LINKED ISSUE #276: Show-InstallationRestartPrompt's [Restart Later] button does nothing
If you call the Countdown-to-Reboot GUI from inside Deploy-Application.ps1:

`Show-InstallationRestartPrompt -Countdownseconds 600`

This calls AppDeployToolkitMain.ps1 a second time to display the Countdown-to-Reboot GUI without blocking Deploy-Application.ps1 that called it.
This works fine but the [Restart Later] button just minimizes the GUI and the countdown timer keeps on going.

This calls AppDeployToolkitMain.ps1 a second time as a separate instance to display the Countdown-to-Reboot GUI without blocking Deploy-Application.ps1 that called it.

I managed to get a debugger going on the second instance that Show-InstallationRestartPrompt creates and this is the problem code:

```
		[scriptblock]$buttonRestartLater_Click = {
			## Minimize the form
			$formRestart.WindowState = 'Minimized'
			## Reset the persistence timer
			$timerPersist.Stop()
			$timerPersist.Start()
		}
```

It turns out for some reason, $timerPersist just stops existing or is just not accessible from the code block.
```
[DBG]: PS> $timerPersist.Stop()
You cannot call a method on a null-valued expression.
At line:1 char:1
+ $timerPersist.Stop()
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull
```

## Solution
```
		[scriptblock]$buttonRestartLater_Click = {
			## Close the form
			$formRestart.Close()     #does nothing
			Start-sleep -Milliseconds 500
			$formRestart.Dispose() #Nuclear option works!
			Start-sleep -Milliseconds 500
		}
```

This works but it does not exit the script as gracefully as I would have wished.
This line causes an error:
`[Microsoft.VisualBasic.Interaction]::AppActivate($powershellProcess.ID)`

For myself, I wrapped it in a try/catch but I think it's a sloppy solution. I'm hoping someone can come up with a better workaround.

luki1412:
This button is misleading for the user. Microsoft uses Restart Later buttons to extend the duration of time left before a reboot is triggered. The function Show-InstallationRestartPrompt however does not extend the duration even though the button is called Restart Later. It only minimizes the window. So this PR renames the button to Minimize. Fixes #276
minimize

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
@@ -188,7 +188,7 @@
 		<RestartPrompt_MessageRestart>Your computer will be automatically restarted at the end of the countdown.</RestartPrompt_MessageRestart>
 		<!-- Text displayed on the restart prompt. -->
 		<RestartPrompt_TimeRemaining>Time remaining:</RestartPrompt_TimeRemaining>
-		<RestartPrompt_ButtonRestartLater>Restart Later</RestartPrompt_ButtonRestartLater>
+		<RestartPrompt_ButtonRestartLater>Minimize</RestartPrompt_ButtonRestartLater>
 		<RestartPrompt_ButtonRestartNow>Restart Now</RestartPrompt_ButtonRestartNow>
 	</UI_Messages_EN>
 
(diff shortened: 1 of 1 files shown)


sintaxasn:
Can’t fault that logic :)

sintaxasn:
Can’t fault that logic :)

PR #528, opened 2020-07-03
Remove-Folder without recursion aka remove only empty folder

psour:

        I think the change now, does what is intended by the function, no change neccessary.

- [X] The enconding of the file wasn't changed. It is still UTF8 without BOM.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -3531,8 +3531,12 @@ Function Remove-Folder {
 			If (Test-Path -LiteralPath $Path -PathType 'Container') {
 				Try {
 					If ($DisableRecursion) {
-						Write-Log -Message "Delete folder [$path] without recursion..." -Source ${CmdletName}
-						Remove-Item -LiteralPath $Path -Force -ErrorAction 'SilentlyContinue' -ErrorVariable '+ErrorRemoveFolder'
+                        if((get-childitem -LiteralPath $Path -ErrorAction 'SilentlyContinue' -ErrorVariable  '+ErrorRemoveFolder').Count -eq 0){
+						    Write-Log -Message "Delete folder [$path] without recursion..." -Source ${CmdletName}
+                            Remove-Item -LiteralPath $Path -Force -ErrorAction 'SilentlyContinue' -ErrorVariable '+ErrorRemoveFolder'
+                        } else {
+                            Write-Log -Message "Delete folder [$path] not possible Folder not empty..." -Source ${CmdletName}
+                        }
 					} else {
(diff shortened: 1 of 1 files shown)


psour:
First Pull request ever, please be kind^^

luki1412:
> First Pull request ever, please be kind^^

First of all, thanks for the pull request. No worries. We have no issue with people adding PRs. I'm just not sure what this one is supposed achieve.

This function is for removing folders. You are changing it to a function that without recursion doesn't remove anything other than an empty folder. I don't see an issue with the code just with the idea.

Instead, you could modify it to accept a parameter -RemoveOnlyIfEmpty and if it is was specified, then the function would remove folders only if they are empty. Should work with and without recursion. 

Just my opinion, maybe other devs have a different opinion. Not trying to be mean. 
Keep the requests coming :)

RobinBeismann:
I do see the need for a switch which removes the folders recursively incase they are empty, that would make sense for a lot of unclean uninstallers, but yeah, this should be a switch not a replacement.

psour:
> 
> 
> I do see the need for a switch which removes the folders recursively incase they are empty, that would make sense for a lot of unclean uninstallers, but yeah, this should be a switch not a replacement.

Maybe I don't get the Switch -DisableRecursion, but what should it do in first place? Because you get a challange response, if the folder you try to delete is not empty. The change works around this challenge response if the foder is not empty. But if the switch should only delete folders without folders in it but having files then it should be changed anyway.

Maybe I should have added an issue with the PR to be more clear.

luki1412:
> > I do see the need for a switch which removes the folders recursively incase they are empty, that would make sense for a lot of unclean uninstallers, but yeah, this should be a switch not a replacement.
> 
> Maybe I don't get the Switch -DisableRecursion, but what should it do in first place? Because you get a challange response, if the folder you try to delete is not empty. The change works around this challenge response if the foder is not empty. But if the switch should only delete folders without folders in it but having files then it should be changed anyway.
> 
> Maybe I should have added an issue with the PR to be more clear.
> 
> > This function is for removing folders. You are changing it to a function that without recursion doesn't remove anything other than an empty folder. I don't see an issue with the code just with the idea.
> 
> But this is not what the switch disablerecursion is doing. It is not working if the folder is not empty, because of challenge response. I added simply a check to delete only empty directories. But I could be wrong with my assumption, that disablerecursion should disable recursion.

Remove folder function is for removing a folder you specify. Recursion means that if there are folders within this folder, the function will delete them also and folders and files within those folders and so on ... Without recursion, you are only removing files directly inside that folder.

psour:
Ah ok, then it is silly how I changed the func. 
Can I take back the Pull Request and send a new on? So the -DisableRecursion Switch does not end up in a Challenge Response, if the folder contains subfolders? Is it simply "close and comment"? I'll test...

luki1412:
Yes you can close your own PR if you want.

PR #534, opened 2020-07-21
Do not depend on PATH for system .exe paths

luki1412:
If the PATH variable is modified and no longer contains path to System32, the toolkit doesnt work correctly. This fix changes those paths so they dont rely on PATH.


sintaxasn:
Nice job :)

PR #535, opened 2020-07-22
Use full registry paths instead of PSdrives

luki1412:
Some of the registry paths used by the toolkit are still using powershell drives that are akin to command aliases and can be changed. This PR changes them to full paths.


sintaxasn:
I don't they'll ever be deprecated but still, i'm in agreement to change.

PR #537, opened 2020-07-24
Add process object to Get-WindowTitle

cybermoloch:
It is useful to have the full object returned for builtin functions such
as CloseMainWindow()

- [X] The enconding of the file wasn't changed. It is still UTF8 without BOM.


cybermoloch:
I think in the future the current ParentProcess should be called ParentProcessName to prevent confusion.

It would also make sense in the future to change the function name to just Get-Window as it returns parameters for the window and not just the title. (And the default parameter should be Name as well...)

PR #540, opened 2020-07-31
Change Un/Block-AppExecution so it doesn't run without admin rights

### LINKED ISSUE #526: -BlockedExecution pop-up does not show when Toolkit_TempPath is set to $envTemp
**Describe the bug**
Due to the fixes made for the security issue in 3.8.2 [Reserved CVE-2020-10962],  the BlockedExecution-Window will not open, when a User tries to open a blocked app. Instead you will see an Error-Message from Windows Script Host.

**To Reproduce**
1. Deployment-Script with -BlockedExecution and -CloseApps in Show-InstallationWelcome 
2. Open one of the blocked apps as user without admin rights
3. Run the Deploy-Script with admin rights
3. Close the blocked app, when prompt asks you to do
4. Try to reopen the blocked app as user without admin rights, while the script is still running
5. you will get an error message from Windows Script Host
6. Change <Toolkit_TempPath> in 'AppDeployToolkitConfig.xml' to $envSystemDrive\Users\Public
7. Redo Steps 2 - 4
8. Now the pop-up will be shown

**Toolkit Version:**
3.8.2

luki1412:
Since $envTemp points to c:\windows\temp and the folder has limited permissions for Administrators by default, it has to be changed again. ProgramData seems like the best option other than ProgramFiles which is used for programs themselves. ProgramData is also hidden by default.

ProgramData is read only to users by default so we need to change the no admin rights log dir to a user writable location for when the toolkit runs without admin rights. Not many such folders exist. One of the best locations is LocalAppData since those folders dont get erased without reason and are user writable.

Also modifies Block and Unblock AppExecution functions so they get bypassed if admin rights are not requested because the functions require elevated permissions to function.

Fixes #526

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
@@ -17,7 +17,7 @@
 		<Toolkit_RequireAdmin>True</Toolkit_RequireAdmin>
 		<!-- Specify if Administrator Rights are required. NB: Some functions won't work if this is set to false, such as deferral, blockexecution, file & registry RW access and potentially logging. -->
 		<Toolkit_TempPath>$envTemp</Toolkit_TempPath>
-		<!-- Path used to store temporary Toolkit files (with PSAppDeployToolkit as subdirectory), e.g. cache toolkit for cleaning up blocked apps. Normally you don't want this set to a path that is writable by regular users, this might lead to a security vulnerability. The default Temp variable for the LocalSystem account is C:\ProgramData. -->
+		<!-- Path used to store temporary Toolkit files (with PSAppDeployToolkit as subdirectory), e.g. cache toolkit for cleaning up blocked apps. Normally you don't want this set to a path that is writable by regular users, this might lead to a security vulnerability. The default Temp variable for the LocalSystem account is C:\Windows\Temp. -->
 		<Toolkit_RegPath>HKLM:\SOFTWARE</Toolkit_RegPath>
 		<!-- Registry key used to store toolkit information (with PSAppDeployToolkit as child registry key), e.g. deferral history. -->
 		<Toolkit_LogPath>$envWinDir\Logs\Software</Toolkit_LogPath>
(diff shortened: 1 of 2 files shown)


cybermoloch:
`ProgramData` is a special folder but is actually writable by users. From [Microsoft](https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-shell-setup-folderlocations-programdata
):
*"ProgramData specifies the path to the program-data folder (normally C:\ProgramData). Unlike the Program Files folder, this folder can be used by applications to store data for standard users, because it does not require elevated permissions."*

It is special because they don't have all the permissions over it for writing but `CREATE` permissions and that grants them additional permissions based on inherited permissions that apply as `CREATOR OWNER`. It seems like they may not have write access because if it was created by say, `SYSTEM`, they cannot write to that file. (At least that is my current understanding.)

(Not sure how that information effects the PR though.)

luki1412:
> `ProgramData` is a special folder but is actually writable by users. From [Microsoft](https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-shell-setup-folderlocations-programdata):
> _"ProgramData specifies the path to the program-data folder (normally C:\ProgramData). Unlike the Program Files folder, this folder can be used by applications to store data for standard users, because it does not require elevated permissions."_
> 
> It is special because they don't have all the permissions over it for writing but `CREATE` permissions and that grants them additional permissions based on inherited permissions that apply as `CREATOR OWNER`. It seems like they may not have write access because if it was created by say, `SYSTEM`, they cannot write to that file. (At least that is my current understanding.)
> 
> (Not sure how that information effects the PR though.)

Yeah but considering these toolkits can be used for user deployments, the logs and files would get overwritten if we use the same folder and you have a deployment for two users on the same machine.

cybermoloch:
I am not sure what the best path is, just wanted to mention that is was kind of user writable.

sintaxasn:
So, have we got a reported issue for this? These paths were set for over 7 years, and only changed recently to remediate a reported CVE. This was a small change to implement, but it's a big deal to a company with 120,000 endpoints - it may be difficult, time consuming or costly to advise or retrain support staff in multiple countries with different languages,  who rely on the log file for troubleshooting installation issues.

Given the roadmap changes we discussed - can this be tabled until we make the larger overall change?

luki1412:
> So, have we got a reported issue for this? These paths were set for over 7 years, and only changed recently to remediate a reported CVE. This was a small change to implement, but it's a big deal to a company with 120,000 endpoints - it may be difficult, time consuming or costly to advise or retrain support staff in multiple countries with different languages, who rely on the log file for troubleshooting installation issues.
> 
> Given the roadmap changes we discussed - can this be tabled until we make the larger overall change?

Well the user variables are new so I dont think those are used at all.
The admin temp path needs to be changed because we changed it recently but the rights are very limited for Windows Temp and its causing issues. Can look at the linked issue or the forums. (https://discourse.psappdeploytoolkit.com/t/loading-script-blockappexecutionmessage-vbs-failed-access-is-denied/2723/2)

luki1412:
Removed path changes for this PR. They will be implemented in another PR

PR #545, opened 2020-08-11
Fix for tempory folder issue for blocked execution

cybermoloch:
Set READ permissions to BUITLTIN\Users so they can get the popup message
Added Set-Permission function to make it easier to set permissions

- [TODO] I updated the documentation with the changes I made.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,4 +6,4 @@ Before submitting this PR, I made sure:
 
 - [X] The code I changed has comments with explanation.
 
-- [X] The enconding of the file wasn't changed. It is still UTF8 without BOM.
+- [X] The encoding of the file wasn't changed. It is still UTF8 without BOM.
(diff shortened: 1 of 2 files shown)


cybermoloch:
Set-Permission function was recommended in an issue and originally from here https://github.com/razy69/Powershell/blob/master/Release/Set-Permission.ps1

Confirmed with original author that it can be used without restriction. "If you keep the code as it is, please just mention my name as author. Otherwise do as you want." It was minimally altered for logging, etc. Further enhancements to the function can be done in the future once this issue is fixed.

Regarding the Execution Block errors, this is because there is a scheduled task made that gets run in the user context to show that the app was blocked. Since the location the task uses is unreadable by a normal user, a permission denied message is popped. Only read rights are required however. Also, changing back the temporary folder to a user writable location when using the toolkit with elevated rights could reintroduce the CVE (as far as I understand -- I have not seen the CVE details).

I have tested this on Windows 10 but I don't think anything in this function won't work in PowerShell 2.

luki1412:
Fix for #526

cybermoloch:
Checked now with PowerShell 2. Will do more testing in different scenarios but it should be good to go.

PR #556, opened 2020-09-17

Basti890:
Change timestamp of compressed ToolikitLogs to 24h format to avoid possible double filenames


sintaxasn (approved):
Approved and thank you for contributing :) Nice catch on this btw.

PR #559, opened 2020-09-28
Set-Itempermission fixes and improvements

luki1412:
Reworked the function, added new parameters.

Also added a couple of missing -Source parameters on some Write-Log uses


sintaxasn (approved):
Great work Luki! :)

PR #560, opened 2020-09-28
Change Get-ScheduledTask and add an alias

### LINKED ISSUE #373: Get-ScheduledTask Conflict
The Get-ScheduledTask function in the PSAppDeployToolkit conflicts with the native powershell cmdlet Get-ScheduledTask. Cannot modify a scheduled task because of the conflict. Can this function be renamed in a future update?

### LINKED ISSUE #558: Get-ScheduledTask doesn't work properly
**Describe the bug**
If you are using the cmdlet "Get-ScheduledTask" it will fail in certain scenarios due to the fact that the main toolkit has a custom function ALSO called "Get-ScheduledTask" that is not as feature rich as the built-in cmdlet.

**To Reproduce**

I noticed the issue when I had the below code block in the Installation section of the script. At first it found an additional task that also had the term "deployment" in it, but as a small part of the full name of a built-in task. I resolved that and then it didn't recognize the "endboundary" property. This code block works fine when run outside of the Deploy-Application.ps1 because it isn't using the custom function in the PSAppDeployTookit. The error message confused me since there was NO indication it was using a custom function with the same exact name as a built-in function. I fixed the issue by renaming the custom "Get-ScheduledTask" function and all references to it.

```
        $run = (Get-Date).AddMinutes(-1).ToString('s')
        $deploytask = Get-ScheduledTask -TaskName "Deployment"
        $deploytask.Triggers[0].endboundary = $run
        $deploytask | Set-ScheduledTask

```

**Screenshots**

**Toolkit Version:**
3.8.2

**Powershell Version:**
5.1

**Additional context**

luki1412:
Renames Get-ScheduledTask to Get-SchedulerTask and adds an alias Get-ScheduledTask if no such function exists.

Fixes #373 
Fixes #558

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -5444,8 +5444,8 @@ Set-Alias -Name 'Refresh-SessionEnvironmentVariables' -Value 'Update-SessionEnvi
 #endregion
 
 
-#region Function Get-ScheduledTask
-Function Get-ScheduledTask {
+#region Function Get-SchedulerTask
+Function Get-SchedulerTask {
 <#
 .SYNOPSIS
 	Retrieve all details for scheduled tasks on the local computer.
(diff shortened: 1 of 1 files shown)


sintaxasn (approved):
Nice job

headofire:
This is fantastic, thank you!

PR #581, opened 2020-10-29
Fix Boolean parameters not being passed to Execute-Process

### LINKED ISSUE #580: Error in function Execute-MSI and parameter ExitOnProcessFailure
**Describe the bug**

The ExitOnProcessFailure parameter does not work with `Execute-MSI`. If you set the parameter to `$false`, then this parameter will not be passed on to `Execute-Process` function.

https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/5706e1eec6750ff4c1194589012b07b14ccbb274/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1#L2660

With `$false` the parameter is not added to the splatting object. Since the function `Execute-Process` sets the parameter to `$true` by default, 
https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/5706e1eec6750ff4c1194589012b07b14ccbb274/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1#L3023
the function will always execute the `Exit-Script` in the event of an error. The same applies to all [Boolean] and [switch] parameters that are transferred to the `Execute-Process` function from `Execute-MSI`.

**To Reproduce**

`$ExecutionMsiResult = Execute-MSI -Action "Install" -Path "DummyMSI-1603.msi" -Transform "DummyMSI-1603.mst" -ExitOnProcessFailure $false -PassThru`

**Screenshots**

**Toolkit Version:**
Version of our toolkit you are using. 3.8.2

**Powershell Version:**
```
Name                           Value
----                           -----
PSVersion                      5.1.14409.1018
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14409.1018
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

**Additional context**

ducke:
Change the behavior of the function Execute-Process and Execute-Msi to check boolean and switch parameters with $PSBoundParameters.ContainsKey()

Fix for #580

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,5 @@
 **Version 3.8.4 [??/??/2020]**
+- Fix Boolean parameters not being passed to Execute-Process
 - Changed Show-InstallationWelcome:
   - The buttons are now wider
   - The listbox is also wider so it is aligned with buttons
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -2652,13 +2652,13 @@ Function Execute-MSI {
 												  Parameters = $argsMSI
 												  WindowStyle = 'Normal' }
 			If ($WorkingDirectory) { $ExecuteProcessSplat.Add( 'WorkingDirectory', $WorkingDirectory) }
-			If ($ContinueOnError) { $ExecuteProcessSplat.Add( 'ContinueOnError', $ContinueOnError) }
+			If ($PSBoundParameters.ContainsKey("ContinueOnError")) { $ExecuteProcessSplat.Add( 'ContinueOnError', $ContinueOnError) }
 			If ($SecureParameters) { $ExecuteProcessSplat.Add( 'SecureParameters', $SecureParameters) }
-			If ($PassThru) { $ExecuteProcessSplat.Add( 'PassThru', $PassThru) }
+			If ($PSBoundParameters.ContainsKey("PassThru")) { $ExecuteProcessSplat.Add( 'PassThru', $PassThru) }
 			If ($IgnoreExitCodes) {  $ExecuteProcessSplat.Add( 'IgnoreExitCodes', $IgnoreExitCodes) }
 			If ($PriorityClass) {  $ExecuteProcessSplat.Add( 'PriorityClass', $PriorityClass) }
-			If ($ExitOnProcessFailure) {  $ExecuteProcessSplat.Add( 'ExitOnProcessFailure', $ExitOnProcessFailure) }
-			If ($NoWait) { $ExecuteProcessSplat.Add( 'NoWait', $NoWait) }
+			If ($PSBoundParameters.ContainsKey("ExitOnProcessFailure")) {  $ExecuteProcessSplat.Add( 'ExitOnProcessFailure', $ExitOnProcessFailure) }
+			If ($PSBoundParameters.ContainsKey("NoWait")) { $ExecuteProcessSplat.Add( 'NoWait', $NoWait) }
 			#  Call the Execute-Process function
 			If ($PassThru) {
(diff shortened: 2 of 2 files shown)


luki1412:
Please read my comment on your issue and modify the PR accordingly.

ducke:
Thanks for your comment.
I've added all switch and boolean parameters directly to the Splat Objects

luki1412:
Again, this only concerns Booleans, not Switches. They work correctly already.

ducke:
Thx for your input. i have put a small demo together, that shows the difference approche regarding splatting and boolean and switches.
https://gist.github.com/ducke/012652246c586ec0b004bcbaaabe20e2
The first code block is the current implementation
`Execute-Msi -exeMsiexec "Foobar.msi" -ExitOnProcessFailure $false`
Output from Execute-Process
`Path=Foobar.msi Parameters= WindowStyle=Normal PassThru=False SecureParameters=False ExitOnProcessFailure=True ContinueOnError=False NoWait=False`

The second code block is your proposed change
`Invoke-Msi -exeMsiexec "Foobar.msi" -ExitOnProcessFailure $false`
Output
`Path=Foobar.msi Parameters= WindowStyle=Normal PassThru=False SecureParameters=False ExitOnProcessFailure=False ContinueOnError=False NoWait=False`

The third code block represents the change in the pull request. All switch and boolean are together in the splatting obejct
`Invoke-SplatMsiExec -exeMsiexec "Foobar.msi" -ExitOnProcessFailure $false`
Output 
`Path=Foobar.msi Parameters= WindowStyle=Normal PassThru=False SecureParameters=False ExitOnProcessFailure=False ContinueOnError=False NoWait=False`

Both works perfect. In the last one are the if statements not necessary.

Once again sorry for asking. i just want to understand but i realize that you only want small changes to the code

luki1412:
Yes the main reason is that we want try to avoid big changes to code. The second reason is that you are just passing variables containing $null to the hash table for no reason. So the hash table still contains a property called  'PassThru' but its value is $null so Powershell ignores it when going through all the properties while parsing parameters.

ducke:
But [switch] parameter are never $null. They are either true or false.

```
function a {
    param (
        [switch]$PassThru,
        [boolean]$ContinueOnError
    )

    $Splat = @{
        PassThru = $PassThru
        ContinueOnError = $ContinueOnError
    }
    $Splat.PassThru

    b @Splat
}
function b {
    param (
        [switch]$PassThru,
        [boolean]$ContinueOnError
    )

    "PassThru={0}, ContinueOnError={1}" -f $PassThru,$ContinueOnError

}

a -ContinueOnError $True
```

luki1412:
Again switch parameter is not the same as Bool: https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.switchparameter

It does have a ToBool() function which converts its value to a Bool, which is what I assume you get when you call ToString()

ducke:
Thanks for taking the time to answer my weird questions :+1: 

I already know that switch and boolean are not the same.
But according to your logic you are not allowed to set the default value of a switch parameter to `$false` in the param block.

```powershell
[Parameter(Mandatory=$false)]
[ValidateNotNullorEmpty()]
[switch]$PassThru = $false,
```

The fact that Powershell turns a switch parameter into `$true` or `$false` at the end of the day is great. Then the developer no longer has to worry about the conversion :smile: 

So be it, if you still want, I will change my pull request for the functions Execute-Msi and Remove-MSIApplications.
ContinueOnError and ExitOnProcessFailure come into the splatting object and I remove the associated IF query.

luki1412:
No I'm just trying to say that most people think that switches and bools are the same thing. They are not, Powershell just makes it look that way. It's very misleading. My original point was that switches were not broken, but you are changing those as well for no reason, which still stands. So please revert them back and add an entry to the changelog.txt - the title of this PR.

Thank you for the PR and the Bug report

luki1412 (approved):

PR #586, opened 2020-11-21
Improved the computing of $dpiScale

denisrennes:
Improved the computing of $dpiScale in the scriptblok $GetDisplayScaleFactor in AppDeployToolkitMain.ps1

Now $dpiScale is more accurate: any value of scaling will be taken into account, included 175% or 300% that will no longer default to 100.
Examples:
- For a user having set a scaling to 175%, $dpiPixels is 168. The computed $dpiScale is now 175 instead of 100.
- For a 4K big screen $dpiPixels is 288 when the scaling is set to "300% (Recommended)". The computed $dpiScale is now 300 instead of 100.

- [  ] I did not update the documentation with the changes I made because $dpiScale is not in the doc, it is only written to the log.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -531,13 +531,7 @@ If (Test-Path -LiteralPath 'variable:deferDays') { Remove-Variable -Name 'deferD
 		[int32]$dpiPixels = Get-RegistryKey -Key 'Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\FontDPI' -Value 'LogPixels'
 		[boolean]$UserDisplayScaleFactor = $false
 	}
-	Switch ($dpiPixels) {
-		96 { [int32]$dpiScale = 100 }
-		120 { [int32]$dpiScale = 125 }
-		144 { [int32]$dpiScale = 150 }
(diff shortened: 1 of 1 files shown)


luki1412:
Although you fixed the calculation, the code above it, is flawed. It almost always returns 96 DPI. We need to fix that before fixing the calculation of the scale makes any sense. The variable isnt used inside our toolkit anywhere anymore but as a variable it is not useful since its almost always equal to 100.

denisrennes:
The name $dpiPixels is a bit misleading: Actually $dpiPixels and $dpiScale both reflect the display scale factor. Yes the display scale is almost always 100% (96 for $dpiPixel) because for most screens, the screen size and resolution are balanced, within "normal" size and resolution ranges, so the default display scale is set to 100%. (The hardware probably provides its size and resolution to the OS).
The default display scale will be higher only for high-density displays, such as a Microsoft Surface Pro with a display resolution of 2736 x 1824 and a 12.3" screen size: 200% by default because at 100%, the displayed text would be too small.
The display scale will sometimes be higher for a more "normal" density display, when the user has changed the scale because he/she is more comfortable with bigger characters. For example, he/she puts 125% for a 17" screen with a resolution of 1920 x 1080. 

However, I realized that there might be a problem with my change: when no user is logged in, maybe $dpiPixel is not set. The current source code defaults to  100 for $dpiScale, I thing I will have to do the same.

Therefore do not validate this code, wait for my next version.

PR #597, opened 2021-01-05
Multiple fixes

luki1412:
A lot of performance improvements in this one.

Changed some function signatures to the ones specified in the official documentation.

Removed some redundant code.

Only loading assemblies once.

Execute-ProcessAsUser now escapes XML characters prior to creating an xml file.

Added log messages to some Show- functions when the function is being bypassed due to deploy mode.

Set-ActiveSetup now doesn't execute the StubPath if the registry key for it exists in the CU hive with IsInstalled set to 1.

Show-InstallationWelcome now closes processes correctly if the user has reopened them while the window was running.

Fixes a couple of typos.

Fixes an issue with Set-ItemPermission where SID translation did not work correctly #598

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,4 @@
-**Version 3.8.4 [??/??/2020]**
+**Version 3.8.4 [??/??/2021]**
 - Fix Boolean parameters not being passed to Execute-Process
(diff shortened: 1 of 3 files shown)


sintaxasn (approved):
Sorry for the delay, took a bit of time to review all the changes! :)

PR #601, opened 2021-01-13
Patch example docu

hbaes:
Just changed documentation block in wiki file. 
For examples it's better to use MD code block 


sintaxasn (approved):
Thanks for this :)

hbaes:
welcome, will rework even more as i have time ;)

PR #609, opened 2021-01-28

dianneman:
Fixed switch handling otherwise this function does not work when dot sourced.


sintaxasn (approved):
Tested and this looks good. Thank you!

sintaxasn:
> Fixed switch handling otherwise this function does not work when dot sourced.

So I can see this fixed the issue but I'm not exactly sure why. Do you mind explaining? I want to be sure we haven't introduced the same issue elsewhere, and that we are aware of the why, so we can avoid it in future. Thanks!

dianneman:
No problem I should have done that. 
The object you use in switch will be the new current object and this object will only contain the types like int32,boolean because of .Value.GetType().Name

_PSItem
You can use the $PSItem or $_ to reference the current item that was processed. When we do a simple match, $PSItem is the value that we are matching. I'll be performing some advanced matches in the next section where this variable is used.
https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-switch?view=powershell-7.1_

rmoreas:
Any idea when v3.8.5 will be released?

wbatzle:
Is this fix going to be published? Looking to resolve this in our environment.

PR #621, opened 2021-02-09
Show-installationprompt function did not work when used with -nowait parameter in combination with boolean arguments

dianneman:


luki1412:
**Thank you** for this pull request, however we have found additional bugs inside the function, so we will include your changes in a bigger pull request that will resolve multiple issues.

dianneman:
Good job! Thanks for the updates :)

Issue #638, opened 2021-03-27
Function to check whether a user or computer is member of a specific Active Directory group without using the Active Directory PowerShell module


powershellking:
**Is your feature request related to a problem? Please describe.**
In some cases it's necessary to be able to check in which Active Directory groups a user or computer is member.

**Describe the solution you'd like**
I'd like to have a function that checks whether a user or computer is member in a specific active directory group. The function must not rely on the Active Directory module of PowerShell since it's not available on every machine.

**Describe alternatives you've considered**
I've already created such a function and added it to the template of our company's version of PSADT. I just wanted to share it with the community so everyone can use it. I'll soon send a pull request on that.

mjr4077au:
Unfortunately the PR you've submitted is too old to consider merging, however I also don't think this functionality needs to be in the toolkit as you shouldn't be deploying to all users/computers, then doing deterministic actions based on group membership at runtime; rather you should configure SCCM or Intune to deploy to your targeted users/computers instead.

PR #639, opened 2021-03-27
Added function Test-ADGroupMembership

powershellking:

This pull request contains the function requested in #638

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -11924,6 +11924,369 @@ Function Set-ItemPermission {
 }
 #endregion
 
+#region Function Test-ADGroupMembership
+function Test-ADGroupMembership
+{
+	<#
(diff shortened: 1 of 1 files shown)


luki1412:
Is there any benefit of using this method over https://docs.microsoft.com/en-us/dotnet/api/system.directoryservices.directorysearcher

powershellking:
As I understand, when using the directory searcher you'll have to specify a LDAP path where to search. The way I implemented it doesn't require specifying a search root so you don't have to care about in which OU the user and group object are located.

luki1412:
> As I understand, when using the directory searcher you'll have to specify a LDAP path where to search. The way I implemented it doesn't require specifying a search root so you don't have to care about in which OU the user and group object are located.

No, you only need a filter. It also supports older net framework versions. It can return a lot of data about the user you specify, unless you only tell it to select certain properties so it could also serve as AD data retrieval function.

seanlillis:
The team have reviewed this pull request and have decided to reject it as this class is known to be buggy.

PR #653, opened 2021-05-11
Add 'vomus' instead of 'v' when $RepairFromSource = $true

DanGough:
Since `/f` implies `/fomus`, `/fv` effectively strips away _omus_, so nothing is actually repaired.
`$RepairFromSource = $true` now uses `/fvomus`.

(I have been using this change in production already)

(no changes required, docs do not go into detail about implementation)

(comments added to commits but not code)


sintaxasn (approved):
Nice catch, thank you. 

@mmashwani, let's get this included when you've submitted your updates.

PR #659, opened 2021-06-16
Update Show-InstallationRestartPrompt.md

hkystar35:
Updated to include new parameters introduced in 3.8.3


sintaxasn (approved):
Thanks for doing this :)

PR #694, opened 2022-02-18
Add new WindowLocations (Top, TopRight, TopLeft, Bottom, BottomRight)

grandweb4:
I'm frustrated when I try to change the window location and PSATK doesn't have options such as "BottomLeft" but it has "BottomRight" so I added the functionality.

- [x] I tested the toolkit with my changes and made sure it doesn't break other code.


DanGough (commented):
I have checked these changes over and would be happy to include them.

DanGough (approved):
I have checked these changes over and would be happy to include them.

DanGough:
Other changes have been contributed into dev that conflict here, this will need further changes before it can be merged.

PR #697, opened 2022-02-22
Add $IsMultiSessionOS Toolkit Variable

KingLinkTiger:

* [x] I tested the toolkit with my changes and made sure it doesn't break other code.
* [x] I updated the documentation with the changes I made.
* [x] The code I changed has comments with explanation.
* [x] The encoding of the file wasn't changed. It is still UTF8 with BOM.

Add $IsMultiSessionOS Toolkit Variable. This allows users to easily check if the current operating system is Windows Multi-Session or not. Keying off of $envOSName at guidance from a coworker and Microsoft.

PS Sorry for PR #695.


seanlillis:
This will be merged in release v3.9.1

PR #735, opened 2023-01-13
Execute-ProcessAsUser VBScript that will start PowerShell fix

bshiznit:
Be sure to also set EscapedPath and EscapedParameters in the code branch that creates a .vbs in the event you are trying to run PowerShell as $RunAsActiveUser - as these newer variables are referenced further down during scheduled task creatiion

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -7481,6 +7481,8 @@ https://psappdeploytoolkit.com
             $executeProcessAsUserScript | Out-File -FilePath "$executeAsUserTempPath\$($schTaskName).vbs" -Force -Encoding 'Default' -ErrorAction 'SilentlyContinue'
             $Path = "$envWinDir\System32\wscript.exe"
             $Parameters = "`"$executeAsUserTempPath\$($schTaskName).vbs`""
+            $EscapedPath = $Path
+            $EscapedParameters = $Parameters
 
             Try {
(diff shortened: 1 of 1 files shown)


Pollewops (approved):
I did test those two entries in my script and after making them, the process in completely hidden again. Seems to work.

IerielStanescuOpensource:
Hi there, looks like the new version triggers detection alerts because of the following commandline:

`"C:\WINDOWS\System32\wscript.exe" "C:\Users\xxxxx\AppData\Local\Temp\PSAppDeployToolkit\ExecuteAsUser\PSAppDeployToolkit-ExecuteAsUser.vbs"`

This would be the description of the block reason:
`A script launched with a filename, path, and/or arguments associated with known adware. If this activity is unexpected, review the file and investigate the host for other signs of adware.`

Any thoughts?

AlexDeep6:
> Hi there, looks like the new version triggers detection alerts because of the following commandline:
> 
> `"C:\WINDOWS\System32\wscript.exe" "C:\Users\xxxxx\AppData\Local\Temp\PSAppDeployToolkit\ExecuteAsUser\PSAppDeployToolkit-ExecuteAsUser.vbs"`
> 
> This would be the description of the block reason: `A script launched with a filename, path, and/or arguments associated with known adware. If this activity is unexpected, review the file and investigate the host for other signs of adware.`
> 
> Any thoughts?

We have the same issue. We used ExecuteProcessAsUser in previous packages (3.8.4) without getting this Alert. 
While testing 3.9.1 without any customizations, we get this alert too and the execution of wscript was blocked.

PR #797, opened 2023-04-10
Fixes #785

### LINKED ISSUE #785: Show-InstallationRestartPrompt
**My Question/Issue:**
So I am using PSADTK v3.9.0 and when using Show-InstallationRestartPrompt -Countdownseconds -60 -NoSilentRestart $False, it is not working and my device does not restarts. All I get is the following in the logs and the devices just sit there.
"**Triggering restart silently because the deploy mode is set to Noninteractive and NoSilentRestart is disabled. Timeout is set to 5 seconds**"

**What I have tried so far:**
I have googled the issue and only saw posts from a few versions back as I believe this was an issue in v3.8.4. I have not seen anything about this on the internet for the current version. I have also followed the built-in Help as well but still can not get the system to reboot automatically once the application is installed. 

Thank you everyone for this awesome tool and hoping to get some feedback if I am doing something wrong.

dobbleg1000:


seanlillis:
@dobbleg1000 thanks for submitting this pull request. I had already committed the fix before spotting your pull request.

PR #819, opened 2023-06-27
precompiled cs as wdac audit enable breaks ADT

Numidian:
as per wdac issue https://discourse.psappdeploytoolkit.com/t/appdeploytoolkitmain-cs-could-not-be-opened/2962/5

can't relax security note sure if there is a better way to precompile AppDeployToolkitMain.cs, but precompile allows ADT to execute without relaxing the policy


DanGough:
I understand there are already plans to work around this issue. Also the dll should ideally be compiled and signed during a build process rather than checking in the binary as source.

pvwc:
Is there any timeline for when a fix for this issue will be applied ?

Ath3na-UK:
Also just come across this as implementing WDAC, good to know its being looked at.

mjr4077au:
The issue this problem tries to address has been resolved for the upcoming PSADT v4 release. As such, I'll close this; but thank you for your contribution!

PR #844, opened 2023-12-04
Mask Passwords in logs: AppDeployToolkitMain.ps1

AntoineRv:
Mask passwords showed as command line in logs.
Parameters triggered : 
-Password
-Pwd
-Passwords
-Pwds

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -16050,18 +16050,36 @@ If ($deployAppScriptFriendlyName) {
     Write-Log -Message "[$deployAppScriptFriendlyName] script version is [$deployAppScriptVersion]" -Source $appDeployToolkitName
 }
 If ($deployAppScriptParameters) {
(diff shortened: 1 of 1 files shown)


mjr4077au:
Wouldn't it be better on your side to use `[System.Security.SecureString]` objects rather than passing plain-text passwords around in RAM and then trying to mask them out?

AntoineRv:
Hello,


Thanks for reply.
I will try it. But not everyone is knowing this type of object.
It could be well to implement this feature.

DanGough:
This looks like it overwrites the contents of an existing internal toolkit variable $deployAppScriptParameters, which is used elsewhere in the code.

It would be better to implement password masking in a more generic fashion directly inside Write-Log.

AntoineRv:
> This looks like it overwrites the contents of an existing internal toolkit variable $deployAppScriptParameters, which is used elsewhere in the code.
> 
> It would be better to implement password masking in a more generic fashion directly inside Write-Log.

Yes it overwrites the contents of variable $deployAppScriptParameters.
You're right to implement it inside Write-Log.

DanGough:
Ok, will close this for now, as all efforts for new enhancements are going towards 4.0 which will look pretty different to v3.

But we should have a suggestion for this filed under Issues so that it doesn't get forgotten about!

PR #856, opened 2024-01-18
Exclude defaultuser0 from Get-UserProfiles

sabieler:
Invoke-HKCURegistrySettingsForAllUsers called Get-UserProfiles and it returned defaultuser0. It 'interacted' with the defaultuser0 profile during a SCCM OSD task sequence in Post-Installation steps. It unloaded the defaultuser0 profile which caused CloudExperienceHostBroker.exe to reset OOBE and reboot.

[Post-Installation] :: Load the User [COMPUTERNAME\defaultuser0] registry hive in path [HKEY_USERS\S-1-5-21...
[Post-Installation] :: Unload the User [COMPUTERNAME\defaultuser0] registry hive in path [HKEY_USERS\S-1-5-21...

### DIFF
--- a/Sources/BetaToolkit/AppDeployToolkitMain.ps1
+++ b/Sources/BetaToolkit/AppDeployToolkitMain.ps1
@@ -3516,7 +3516,8 @@ Function Get-UserProfiles {
 			ForEach-Object {
 				Get-ItemProperty -LiteralPath $_.PSPath -ErrorAction 'Stop' | Where-Object { ($_.ProfileImagePath) } |
 				Select-Object @{ Label = 'NTAccount'; Expression = { $(ConvertTo-NTAccountOrSID -SID $_.PSChildName).Value } }, @{ Label = 'SID'; Expression = { $_.PSChildName } }, @{ Label = 'ProfilePath'; Expression = { $_.ProfileImagePath } }
-			}
+			} |
+			Where-Object { $_.NTAccount -notmatch 'defaultuser0$'} # This excludes the "defaultuser0" account used by CloudExperienceHostBroker.exe
 			If ($ExcludeSystemProfiles) {
 				[string[]]$SystemProfiles = 'S-1-5-18', 'S-1-5-19', 'S-1-5-20'
 				[psobject[]]$UserProfiles = $UserProfiles | Where-Object { $SystemProfiles -notcontains $_.SID }
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -6480,7 +6480,7 @@ https://psappdeploytoolkit.com
                     Get-ItemProperty -LiteralPath $_.PSPath -ErrorAction 'Stop' | Where-Object { ($_.ProfileImagePath) } |
                         Select-Object @{ Label = 'NTAccount'; Expression = { $(ConvertTo-NTAccountOrSID -SID $_.PSChildName).Value } }, @{ Label = 'SID'; Expression = { $_.PSChildName } }, @{ Label = 'ProfilePath'; Expression = { $_.ProfileImagePath } }
                     } |
-                    Where-Object { $_.NTAccount } # This removes the "defaultuser0" account, which is a Windows 10 bug
+                    Where-Object { $_.NTAccount -notmatch 'defaultuser0$'} # This excludes the "defaultuser0" account used by CloudExperienceHostBroker.exe
             If ($ExcludeSystemProfiles) {
                 [String[]]$SystemProfiles = 'S-1-5-18', 'S-1-5-19', 'S-1-5-20'
                 [PSObject[]]$UserProfiles = $UserProfiles | Where-Object { $SystemProfiles -notcontains $_.SID }


Sources/BetaToolkit/AppDeployToolkitMain.ps1
```diff
@@ -3516,7 +3516,8 @@ Function Get-UserProfiles {
 			ForEach-Object {
 				Get-ItemProperty -LiteralPath $_.PSPath -ErrorAction 'Stop' | Where-Object { ($_.ProfileImagePath) } |
 				Select-Object @{ Label = 'NTAccount'; Expression = { $(ConvertTo-NTAccountOrSID -SID $_.PSChildName).Value } }, @{ Label = 'SID'; Expression = { $_.PSChildName } }, @{ Label = 'ProfilePath'; Expression = { $_.ProfileImagePath } }
-			}
+			} |
+			Where-Object { $_.NTAccount -notmatch 'defaultuser0$'} # This excludes the "defaultuser0" account used by CloudExperienceHostBroker.exe
```
mjr4077au, line 3520: I'd suggest matching against `defaultuser\d+$`, as its index isn't always 0.

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -6480,7 +6480,7 @@ https://psappdeploytoolkit.com
                     Get-ItemProperty -LiteralPath $_.PSPath -ErrorAction 'Stop' | Where-Object { ($_.ProfileImagePath) } |
                         Select-Object @{ Label = 'NTAccount'; Expression = { $(ConvertTo-NTAccountOrSID -SID $_.PSChildName).Value } }, @{ Label = 'SID'; Expression = { $_.PSChildName } }, @{ Label = 'ProfilePath'; Expression = { $_.ProfileImagePath } }
                     } |
-                    Where-Object { $_.NTAccount } # This removes the "defaultuser0" account, which is a Windows 10 bug
+                    Where-Object { $_.NTAccount -notmatch 'defaultuser0$'} # This excludes the "defaultuser0" account used by CloudExperienceHostBroker.exe
```
mjr4077au, line 6483: Same here also.

Sources/BetaToolkit/AppDeployToolkitMain.ps1
sabieler: True, but defaultuser0 is the one used during OOBE.
@seanlillis Do you agree with my suggestion?

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
sabieler: True, but defaultuser0 is the one used during OOBE.
@seanlillis Do you agree with my suggestion?

rdearson: Just a note, this functionality is used by others for its current behaviour.
We utilise PSADT to enforce a range of configurations during Autopilot that arent available via traditional CSP / Settings Catalogue or they might not fit for purpose in their current state via Intune's native featureset.

Eg, a subset of security baseline ADMX based configurations target HKU instead of HKLM, Intune does not enforce the configurations into defaultuser0 profiles (they dont get deleted after OOBE if the logged in users are all standard, raised this with MS recently & have a change request in for it).

We utilise Qualys for our Security Baseline compliance reporting which reports that defaultuser0 is missing the configurations. Utilising PSADT to enforce the configurations to all profiles is a path of least resistance for us as it is quite difficult getting Qualys to modified the control evaluation logic for a number of configurations.

TLDR, this would introduce a breaking change for our (fringe) usage, of course we shouldnt need to leverage PSADT in the first place in the perfect world....


We also do utilise the Get-UserProfiles function as part of a CMApp in our prod task sequence & havent identified this behaviour before.

sabieler:
@rdearson defaultuser0 will be deleted post-deployment. I'm uncertain why configuring this user is necessary given its transient nature.

seanlillis:
@rdearson @sabieler would a compromise be to exclude defaultuser0 only if running in OOBE or would this not address your concern @rdearson?

rdearson:
Appreciate it @sabieler & @seanlillis.

For us specifically, we have an ESP blocking app PSADT win32 app to fill in the "Autopilot / Intune gaps" that we are required to enforce asap.
I believe that stage of the ESP, OOBE/CloudExperienceHostBroker.exe would still be running.

Just to expand on my previous statement regarding defaultuser0 not being deleted post Autopilot. Microsoft Support have confirmed with the Intune Product Group that the expected behavior is that the defaultuser0 user profile is deleted post Autopilot / ESP, only in the event that a local administrator has logged into the machine interactively.

I'm not going to say there is a high percentage of other enterprises which leverage PSADT in a similar scenario that we are using it for here, but we might not be the only ones :)

Can i suggest an additional parameter to the function itself to permit excluding defaultuser* profiles, or a PSADT variable / xml configuration?

I'd offer to provide assistance writing the proposed changes but am quite time poor currently.

sabieler:
@rdearson  @seanlillis 
The defaultuser0 exists immediately after reaching the desktop, but the account itself is disabled (!) and removed from administrators group (!). Can you please confirm and explain your security concern?

rdearson:
Hey @sabieler, my point is more around introducing a breaking change to existing behaviour, our scenario listed is more of an example where it might not be obvious of the implications of the change to existing consumers of PSADT.

Naturally I do agree that our scenario is a very edge case & don't think we need to jump into the scenario itself too much. (I do agree that it shouldn't be needed.....)

PR #864, opened 2024-03-06

AntoineRv:
Minor fix when restart is nedeed command not executed due to char ` 


DanGough:
I haven't tested this but am curious to know why this is needed - I for one have always escaped double quotes inside double quotes by using the backtick, I never even realised you could just double up on the quotes. In the terminal, both of these snippets from the ArgumentList produce the same output:

```ps
"-ExecutionPolicy Bypass -NoProfile -NoLogo -WindowStyle Hidden -Command `"& { Start-Sleep -Seconds $SilentCountdownSeconds; Restart-Computer -Force; }`""

"-ExecutionPolicy Bypass -NoProfile -NoLogo -WindowStyle Hidden -Command ""& { Start-Sleep -Seconds $SilentCountdownSeconds; Restart-Computer -Force; }"""
```

AntoineRv:
Hi,

Actually it doesn't work with `, sleep + restart doesn't run.

mjr4077au:
Agreed with DanGough, there's zero difference between escaping a double quote in a double-quoted string with another double quote, or by using a backtick.

DanGough:
@AntoineRv please could you explain further on how to reproduce the original issue? I'm not sure what you mean by sleep + restart in the context of PSADT.

AntoineRv:
Let it down. I have not enough time to reproduce. Sorry.

PR #883, opened 2024-03-15
- Miscellaneous fixes that I've lumped together. (Targeted to dev)

mjr4077au:
Most of these are bad variable accesses while running under `Set-StrictMode -Version Latest`. I'd implore you to consider enabling this mode within your module by default to allow stricter checks against undefined behaviour.

- `$configToastDisable` should be a bool and not a string.
- Ensure `$HKULanguages` is defined before attempting to access the variable.
  * Such setups breaks PowerShell when running strict compliance mode, such as `Set-StrictMode -Version Latest`.
- Set `$DisableLogging` to `$true` in `Write-Log` if `$LogFileName` is null/hasn't been set up yet.
  * Was flagging either under PowerShell 7, or when strict mode was on as the call to `Out-File` later on was firing before the `$logFile` global was initialised.
- Remove dead code relating to `$formCloseApps`.
  * Variable seems to be a remnant from a previous setup.
  * The remains of this was causing the script to throw under strict compliance mode.
- Repair `$script:notifyIcon` tests.
  * This was throwing under strict compliance mode as the variable is not guaranteed to be there. This check is safer under such situations.
- Ensure `$stdOut` and `$stdErr` are always available on the stack in `Execute-Process`.
  * These variables were only being defined deep within a branch, then the resulting attempt to access was against variables that possibly didn't exist.
- Leverage pre-existing `$culture` global inside `Invoke-RegisterOrUnregisterDLL`.
- Repair bad `$IsMspFile` setup that caused the upcoming branch to fail in strict compliance mode.

 
### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -1790,7 +1790,7 @@ https://psappdeploytoolkit.com
         New-ZipFile -DestinationArchiveDirectoryPath $configToolkitLogDir -DestinationArchiveFileName $DestinationArchiveFileName -SourceDirectory $logTempFolder -RemoveSourceAfterArchiving
     }
 
-    If ($script:notifyIcon) {
+    If (Test-Path -LiteralPath 'variable:notifyIcon') {
         Try {
             $script:notifyIcon.Dispose()
         }
@@ -10640,7 +10640,7 @@ https://psappdeploytoolkit.com
             Return
         }
         ## Dispose of previous balloon
-        If ($script:notifyIcon) {
+        If (Test-Path -LiteralPath 'variable:notifyIcon') {
             Try {
                 $script:notifyIcon.Dispose()
             }
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -1134,6 +1134,9 @@ https://psappdeploytoolkit.com
         If (-not (Test-Path -LiteralPath 'variable:DisableLogging')) {
             $DisableLogging = $false
         }
+        If ([System.String]::IsNullOrWhiteSpace($LogFileName)) {
+            $DisableLogging = $true
+        }
         #  Check if the script section is defined
         [Boolean]$ScriptSectionDefined = [Boolean](-not [String]::IsNullOrEmpty($ScriptSection))
         #  Get the file name of the source script
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -12261,9 +12261,7 @@ https://psappdeploytoolkit.com
             ## Create a Windows Installer object
             [__ComObject]$Installer = New-Object -ComObject 'WindowsInstaller.Installer' -ErrorAction 'Stop'
             ## Determine if the database file is a patch (.msp) or not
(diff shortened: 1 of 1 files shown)


Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -1694,11 +1698,6 @@ https://psappdeploytoolkit.com
     ## Get the name of this function
     [String]${CmdletName} = $PSCmdlet.MyInvocation.MyCommand.Name
 
-    ## Stop the Close Program Dialog if running
-    If ($formCloseApps) {
-        $formCloseApps.Close
```
sintaxasn, line 1699: Did you have a reason for removing this ?

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -1791,7 +1790,7 @@ https://psappdeploytoolkit.com
         New-ZipFile -DestinationArchiveDirectoryPath $configToolkitLogDir -DestinationArchiveFileName $DestinationArchiveFileName -SourceDirectory $logTempFolder -RemoveSourceAfterArchiving
     }
 
-    If ($script:notifyIcon) {
+    If (Test-Path -LiteralPath 'variable:notifyIcon') {
```
sintaxasn, line 1793: What's the purpose behind this change?

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -10641,7 +10641,7 @@ https://psappdeploytoolkit.com
             Return
         }
         ## Dispose of previous balloon
-        If ($script:notifyIcon) {
+        If (Test-Path -LiteralPath 'variable:notifyIcon') {
```
sintaxasn, line 10644: Same question here

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
mjr4077au: It might be best to read the PR descriptions for some of these. I enabled strict compliance mode (`Set-StrictMode -Version Latest`) and started to comb through the errors. This variable is not defined anywhere at all, therefore it's an access on an undefined variable and was causing a throw. It looks like something that's been left behind from a previous setup.

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
mjr4077au: Same here, it's an attempt to access a variable that doesn't always exist depending on whether the toolkit is running silent or not. Under PowerShell's loosey-goosey default mode, it slips through but in strict compliance mode (`Set-StrictMode -Version Latest`), it causes a throw under the right conditions.

Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
mjr4077au: Same answer here as well.

sintaxasn: I think it'd be useful to have a chat about same of these. If you're on Discord, maybe hit me up there?

mjr4077au: Do you have a link at all? The one on the website seemed dead

sintaxasn: 🤦🏻https://discord.gg/winadmins

PR #893, opened 2024-03-16
- Rework all WinForms in PSADT for HiDPI support. (Targeted to dev)

mjr4077au:
* Sets PowerShell process as DPI-aware via a C# P/Invoke.
* Rework all the sizing to use ClientSize properties.
* Rework the auto scaling to be DPI-based, not font-based.
* Clean out bad code in `StateCorrection_Load` delegates that was breaking HiDPI, as well as the forms in PowerShell 7.
* Rework banner sizing to not be hard-coded, but rather factoring in the aspect ratio of the graphic when determining the maximum height.

Before, at 250% DPI on a 4K display:
image

After, at 250% DPI on a 4K display:
image


mjr4077au:
image
Just to prove the PowerShell 7 thing, when used in-conjunction with the other PRs. I'm not saying PowerShell 7 support is 100% (mostly issues with setting up scheduled tasks via `Execute-ProcessAsUser` due to quoting semantics), but we're off to a good start here.

PR #895, opened 2024-03-17
Replace recent append functionality with full log rotation

DanGough:

- [X] I updated the documentation with the changes I made. (only help inside function updated - main docs and site still to do)

I branched off of Master rather than Dev, since that's where the recent commits were made for the 'DoNotAppend' funcionality.
Order of a few params in Write-Log have changed (just for aesthetics) that might technically be a breaking change, although very unlikely people would be using 10+ positional parameters.

- Default is <Toolkit_LogAppend>True</Toolkit_LogAppend> (which behaves the same as 3.9.3)
- Turn that to false and you have managed log rotation. 
- Handling is combined with the max log size routine, which has now moved from the End to the Begin block of Write-Log
- Max history set by <Toolkit_LogMaxHistory>; if set to 0 it will just retain the latest log only.
- Kept the -AppendToLogFile parameter in the function - turns out this is required so that we can force enable append when writing to the end of a log that has reached max size.
- Naming convention to append datetime matches that used by the zip files, with .log extension.
- Previous naming convention to rename to .lo_ when max size reach is also included in the files checked for deletion.
- Same history count is applied to zip files if compression is enabled.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
@@ -44,12 +44,14 @@
         <!-- Specify if log file should be a CMTrace compatible log file or a Legacy text log file. -->
         <Toolkit_LogDebugMessage>False</Toolkit_LogDebugMessage>
         <!-- Specify if debug messages such as bound parameters passed to a function should be logged. -->
-        <Toolkit_LogMaxSize>10</Toolkit_LogMaxSize>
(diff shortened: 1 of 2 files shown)


DanGough:
Hold this for now, I just had a thought - when rotating logs I should use the timestamp of the file for the new name rather than the current time. I'll amend it later.

DanGough:
Line 1243/1244/1813 BTW if you want to change the naming convention from _yyyy-MM-dd-HH-mm-ss (I believe CM default is -yyyyMMdd-HHmmss).

PR #902, opened 2024-03-19
Add log rotation to replace 'DoNotAppend' changes

DanGough:

- [X] I updated the documentation with the changes I made. (only help inside function updated - main docs and site still to do)

- Order of a few params in Write-Log / config.xml have changed (just for aesthetics to group the related params). This might technically be a breaking change, although very unlikely people would be using 10+ positional parameters.
- Default is <Toolkit_LogAppend>True</Toolkit_LogAppend> (which behaves the same as 3.9.3). Turn that to false and you have managed log rotation.
- Handling is combined with the max log size routine, which has now moved from the End to the Begin block of Write-Log.
- Max history set by <Toolkit_LogMaxHistory>; if set to 0 it will just retain the latest log only.
- Kept the -AppendToLogFile parameter in the function - turns out this is required so that we can force enable append when writing to the end of a log that has reached max size.
- Naming convention to append datetime matches that used by the zip files, with .log extension (_yyyy-MM-dd-HH-mm-ss is appended. Note that CM uses -yyyyMMdd-HHmmss which may be preferable).
- Previous naming convention to rename to .lo_ when max size reach is no longer used, but we also include them in the files to be checked for deletion.
- The same history count is applied to zip files if compression is enabled.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
@@ -44,17 +44,19 @@
         <!-- Specify if log file should be a CMTrace compatible log file or a Legacy text log file. -->
         <Toolkit_LogDebugMessage>False</Toolkit_LogDebugMessage>
         <!-- Specify if debug messages such as bound parameters passed to a function should be logged. -->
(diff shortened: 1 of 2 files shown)


sintaxasn (approved):
Looks good

PR #903, opened 2024-03-19
- Re-write `Get-RunningProcesses`.

mjr4077au:
I kept getting property errors while running in strict mode and found the setup in `Get-RunningProcesses` could probably stand to do with a re-write.

Its probably not important enough for 3.9.4, but it does go to show again what `Set-StrictMode` finds that previously would go unnoticed. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/set-strictmode?view=powershell-7.4

image

* Significantly reduce code depth by adopting a "never-nesting" approach. This makes the code more readable by not having to keep track of how many branches down the chain you are.
* Remove almost all else branches by performing early returns where smart to do so. This goes hand in hand as part of never-nesting and keeps the code more readable.
* Speed up code by not looping through `$processObjects` for every single process. The setup in the `Where-Object` filterscript was a bit abusive and also costly.
* Get rid of `Write-Output` usage entirely. It needs to die and is not recommended, or required at all: https://get-powershellblog.blogspot.com/2017/06/lets-kill-write-output.html
* Remove now-unnecessary `Where-Object` call in `Show-InstallationWelcome` due to better setup in `Get-RunningProcesses`. We can just rely on the `Select-Object` call that proceeded it.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -8757,63 +8757,63 @@ https://psappdeploytoolkit.com
         [Switch]$DisableLogging
     )
 
-    Begin {
+    begin {
         ## Get the name of this function and write header
         [String]${CmdletName} = $PSCmdlet.MyInvocation.MyCommand.Name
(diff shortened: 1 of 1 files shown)


sintaxasn (approved):
Great job

PR #906, opened 2024-03-21
- Ensure all UI message strings are appropriately trimmed.

### LINKED ISSUE #900: Certain texts not lining up on WinForms due to indentation within config file.
**Describe the bug**
Because the config file has switched to spaces for indentation, they're being picked up literally in the strings used within the forms. If you look at the "Please save your work text" below, you'll note that it's off-centre.

To address this, you can either go back to tabs, or you can make the string be a single line, then use LF characters to break the line (`&#xA;`).

Benefits of tabs:
- Allows more natural representation of text without users needing to know ASCII control characters.
- More efficient in terms of space and allow user to control indentation depth to their liking.

Benefits of spaces with control character.
- More compatible with editors which seem to hate tabs and default to spaces.
- Using ASCII control characters allows precise control of strings.

EDIT: A 3rd way of addressing this would be to parse each string programmatically on PSADT's side.
For example, change:
`[String]$configClosePromptMessage = $xmlUIMessages.ClosePrompt_Message`
To:
``[String]$configClosePromptMessage = [String]::Join("`n", $xmlUIMessages.ClosePrompt_Message.Split("`n").Trim())``

**Screenshots**
| Current Setup | With Tabs or LF char |
|---------------|----------------------|
| image | image |

**Toolkit Version:**
3.9.4 pre

**Powershell Version:**
Windows PowerShell 5.1

mjr4077au:
* The spaces since the indentation change to the XML are being picked up literally and causing texts to be off-centre.
* This commit splits each string on its new line, trims each line, then joins them back together again.
* Uses System.String static methods instead of PowerShell `-split` and `-join` operators as they're slower.
* Coded solution vs. fixing up the config file covers all bases, including end user deviating configs.
* Fixes #900 


sintaxasn:
Nice

PR #908, opened 2024-03-22
Add -Wow6432Node switches for 32-bit registry support

DanGough:

This adds a -Wow6432Node switch, which is accepted by all the major registry functions as well as Set-ActiveSetup. The value of the switch is passed through to Convert-RegistryPath which does the grunt work of replacing HKLM\Software with HKLM\Software\Wow6432Node (along with a few special edge cases for Classes keys and Active Setup keys under HKCU).

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -5501,6 +5501,10 @@ Converts registry key hives to their full paths. Example: HKLM is converted to "
 
 Path to the registry key to convert (can be a registry hive or fully qualified path)
 
+.PARAMETER Wow6432Node
+
+Specifies that the 32-bit registry view (Wow6432Node) should be used on a 64-bit system.
+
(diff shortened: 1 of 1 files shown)


mjr4077au:
This is a great quick win for sure. I think for v4.0, there should be consideration around rewriting all the registry stuff to use the `[Microsoft.Win32.Registry]` API as you can use RegistryView enums to access the 32-bit registry more effectively.

Below code is shamelessly copied from StackOverflow as I'm on my mobile.

```PowerShell
$key = [Microsoft.Win32.RegistryKey]::OpenBaseKey([Microsoft.Win32.RegistryHive]::LocalMachine, [Microsoft.Win32.RegistryView]::Registry64)
$subKey =  $key.OpenSubKey("SOFTWARE\Microsoft\.NETFramework")
$root = $subKey.GetValue("InstallRoot")
```

DanGough:
Agreed, I looked into that approach also, but it would have required a more major rewrite!

sintaxasn (approved):

PR #918, opened 2024-03-28
`Execute-Process` improvements

mjr4077au:
There's two things included here:
* Allow specification of valid exit codes.
  * This is needed as I'm often running non-MSI installers through PSADT where the measures of success can be anything.
* Add option for `Execute-Process` to be able to throw on failure.
  * I need this so that I can catch a failed execution in `Deploy-Application.ps1`'s main try/catch block as its important I'm able to present the exit code on-screen in a dialog.
  * Being able to throw here could be beneficial to others and removes a bit of boilerplate from my side. 

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -3340,6 +3340,10 @@ Specifies priority class for the process. Options: Idle, Normal, High, AboveNorm
 
 Specifies whether the function should call Exit-Script when the process returns an exit code that is considered an error/failure. Default: $true
 
+.PARAMETER ThrowOnProcessFailure
+
+Specifies whether the function should throw a terminating error when the process returns an exit code that is considered an error/failure. Default: $false
+
 .PARAMETER RepairFromSource
 
 Specifies whether we should repair from source. Also rewrites local cache. Default: $false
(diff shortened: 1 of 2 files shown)


DanGough:
There is already an -IgnoreExitCodes, does this not serve the same purpose as -ValidExitCodes?

mjr4077au:
Upon initial inspection of `Execute-Process`, it didn't seem that `$IgnoreExitCodes` was hooked up. It is, just that the variable is accessed as `$ignoreExitCodes` and I missed it in a case-sensitive search.

I still see value in this. The message returned by an `-IgnoreExitCodes` setup of `"Execution completed and the exit code [$returncode] is being ignored."` is a bit average and doesn't indicate or imply success. This setup also ensures the right stuff in `Exit-Script` is being hit as well, such as success balloon tips and deferral counter cleanups, etc.

Maybe the two parameters can come together, one can be an alias for the other so that we can still get the better ending in `Exit-Script`?

mjr4077au:
Nothing to be sorry about at all! Your point is quite valid, and that little improvement you just mentioned would appease me 50% of the way for sure. Let's leave it marinade a bit and I'll see if I can think of something better for the balloon tip stuff after some shut-eye 🥱

mjr4077au:
@DanGough thanks for the info RE: `-IgnoreExitCodes`. I've reworked my side significantly and I'm using that in conjunction with some other code in Extensions.ps1 to handle what I need.

I still would like some kind of way to specify valid exit codes to `Exit-Script` so I've tacked another commit into my `fixes` branch, for which a PR is already open.

Issue #589, opened 2020-12-07
New Feature: Re-launch toolkit as 64-bit (Intune)


seanlillis:
Intune runs Win32 app installations as 32-bit on 64-bit systems, which means scripts don't write to the correct native registry hives.

Re-launching the toolkit as 64-bit if run in a 32-bit PowerShell process would be a useful feature, like in this example: https://z-nerd.com/blog/2020/03/31-intune-win32-apps-powershell-script-installer/

mjr4077au:
@seanlillis are you still keen on such a feature? I'm personally not a fan as if I want to boot the toolkit up in x86 mode, I should be able to without the toolkit monkeying around on me. People should be using `%SystemRoot%\sysnative\WindowsPowerShell\v1.0\powershell.exe` if they require Intune to use the 64-bit interpreter.

That-Annoying-Guy:
I have code that I put in the Front script to do this.
It forces 64bit and elevation but you can strip out the elevation part out.

mjr4077au:
If you wanted to share it here, feel free to. The other reason why I don't think it's necessary is that the exe handles this for users anyway

That-Annoying-Guy:
While I agree it might not be needed if you use the EXE.
Meanwhile at my day job, we don't use the (v3) EXE because we couldn't make it launch renamed front scripts.
We place this in all our front scripts before the AppDeployToolkitMain.ps1 even loads.

```
# Get the ID and security principal of the current user account
$myWindowsID = [System.Security.Principal.WindowsIdentity]::GetCurrent()
$myWindowsPrincipal = new-object System.Security.Principal.WindowsPrincipal($myWindowsID)

# Get the security principal for the Administrator role
$adminRole = [System.Security.Principal.WindowsBuiltInRole]::Administrator

[Boolean]$Is64BitProcess = [boolean]([IntPtr]::Size -eq 8)

# Check to see if we are currently running "as Administrator" *AND* running 64 bit
If (($myWindowsPrincipal.IsInRole($adminRole)) -and ($Is64BitProcess -eq $true)) {
	# We are running "as Administrator" - so change the title to indicate this
	$Host.UI.RawUI.WindowTitle = $myInvocation.MyCommand.Definition + " (Elevated+64bit) [$DeploymentType]"
	clear-host
} Else {
	If ($Is64BitProcess -eq $false) { Write-Host "NOT running 64Bit" -ForegroundColor Yellow }
[...]

PR #921, opened 2024-03-28
Change default log location based on deployment system

ITsVeritas:
Check if script is executing from ConfigMgr or Intune and if so, change the default log location so that logs can be collected via diagnostics data collection of those respective systems. This option is enabled by default but an additional XML config option has been added to disable this functionality. Proposing to include this as default with the 4.0 release as it can be considered a breaking change. Alternatively, we could set the config file option to False and adjust the documentation changes accordingly though I think this is a nice default feature to include.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitConfig.xml
@@ -40,6 +40,12 @@
         registry key), e.g. deferral history. -->
         <Toolkit_LogPath>$envWinDir\Logs\Software</Toolkit_LogPath>
         <!-- Log path used for Toolkit logging. -->
+        <Toolkit_LogPathConfigMgr>$envWinDir\CCM\Logs</Toolkit_LogPathConfigMgr>
+        <!-- Log path used for Toolkit logging if script path is within CCMCache. -->
+        <Toolkit_LogPathIntune>$envProgramData\Microsoft\IntuneManagementExtension\Logs</Toolkit_LogPathIntune>
+        <!-- Log path used for Toolkit logging if script path is within IMECache. -->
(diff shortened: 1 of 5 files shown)


DanGough:
I think this will be a good enhancement for 4.0. 

4.0 is going to have some fairly major changes though meaning we won't be able to merge this in as-is, but we can certainly take the idea and adapt it for the new codebase!

DanGough:
One other thought - both SCCM and Intune support user context installs, so this would have to be checked to ensure that they still use a log location in the user profile.

mjr4077au:
> If Intune is detected, this known issue should be handle automatically also.
> 
> #589 https://archive.z-nerd.com/blog/2020/03/31-intune-win32-apps-powershell-script-installer

@JonathanPitre I didn't see this comment until just now, but #589 has been closed out as a "won't fix". It's not up to us to decide whether our toolkit should be used via a 64-bit version of PowerShell on every invocation. If I deliberately call powershell.exe and use a 32-bit version, I expect it to run the toolkit in the version of powershell.exe I specify.

In any case, the exe handles this for the user and its our preferred/recommended way to invoke the toolkit. The exe has a `/32` parameter that can be supplied to it for those who wish to use the exe but force a 32-bit version of PowerShell to be used.

DanGough:
@ITsVeritas we are still interested in having this feature in 4.x one day, but I will close this PR now since the commits were for the 3.x codebase which we aren't intending on doing any further releases for.

During initial tests we encountered difficulties reliably detecting if an install had been started by ConfigMgr under all conditions:

- Support for Application and Package models, also if run directly via a Task Sequence
- Whether the application was triggered manually via Software Center or automatically via a required deployment

(You can't always determine the parent process via all of these mechanisms)

- Whether the application was running from ccmcache or from a network share

(Therefore you can't rely on the working dir being set to ccmcache)

PR #924, opened 2024-03-30
Fix for Copy-File

DanGough:

Issue reported on Discord - Copy-File fails if copying a source of `"$dirFiles\*"` with -Recurse, if the source folder contains subfolders but no files in the root.

If a file is specified in the path (or anything with a wildcard), we're supposed to be splitting the path argument so we can give Robocopy SourceFolder, DestFolder and Files.

Previous code only checked to see if the source contains a file via Test-Path -PathType Leaf. Now it also will hit this code branch if the last portion of the path contains a *.


DanGough:
Actually I can probably simplify this even further by just doing the argument splitting if Test-Path -LiteralPath -PathType Container returns false.

I will take another look later this afternoon.

DanGough:
All good to merge now.

DanGough:
Now Flatten with Robocopy supports files/wildcards just like native copy, e.g:

`Copy-File -Source C:\Temp*.txt C:\Temp2 -Flatten`

Also `/XX` (Exclude Xtra) added to Robocopy args to stop it listing all the non-matching files it finds in the destination in the log output.

sintaxasn (approved):

PR #935, opened 2024-04-05
Fix various small formatting errors

Mrman1093:


mjr4077au:
I don't know what the opinion is of the other members of the team, but its probably not worth ruining the git blame for such trivial changes like this.

mjr4077au:
Another thing as well is that you're actually changing lines that are technically correct as well. If you tab-complete `[stri` in a PowerShell window, it'll tab-complete to `[string]`, so the lower-cased iterations are correct and all upper-cased iterations in PSADT are wrong.

mmashwani:
I don't agree that Pascal case type names are incorrect in PSADT. That was a conscious design choice we made at one point to be consistent everywhere type names are used.

Microsoft's design guidelines say to use pascal casing for names spaces and types:
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

If I use Get-Member on a type, the value that is returned is also Pascal cased.

Let say we use [string], and then for some other type, we need to specify the full path.

To use a specific example, I would rather use the following because it is more readable:
[System.Management.Automation.OrderedHashtable]
Than:
[system.management.automation.orderedhashtable]

Or this:
[System.Management.Automation.Cmdlet.ThrowTerminatingError]

is more readable than:
[system.management.automation.cmdlet.throwterminatingerror]

mjr4077au:
Short object names like `[string]` are aliases (with this one being for `[System.String]`, of course. The alias is always lower-case, hence why it tab-completes lower-cased in the console window. When declaring the full name, you would absolutely PascalCase it.

mjr4077au:
@Mrman1093 I'm sorry but there's no way we can accept this now with such invasive changes. I'll close this one for now but we'd welcome other contributions in the future.

Mrman1093:
Sorry about that. I didn't realize that the changes I've been doing to my own repo would be included in a PR that I had already submitted.

mjr4077au:
That's all good mate, I've been caught out before like this. Might I suggest you hold off any large refactoring for now? There's things on the horizon along the lines of where you're heading.

PR #937, opened 2024-04-05
Add -ExcludeServiceProfiles to Get-UserProfiles, add passthru of ExcludeServiceProfiles/ExcludeSystemProfiles/ExcludeDefaultUser/ExcludeNTAccount to Copy-FileToUserProfiles & Remove-FileFromUserProfiles

DanGough:

It makes sense to exclude all service accounts by default (e.g. NT SERVICE\MSSQL$SQLEXPRESS) when getting user profiles. If one could envisage a scenario where this might be a breaking change, the default for ExcludeServiceProfiles in Get-UserProfiles could be changed to $false.

Copy-FileToUserProfiles & Remove-FileFromUserProfiles now pass through and splat ExcludeServiceProfiles/ExcludeSystemProfiles/ExcludeDefaultUser/ExcludeNTAccount to Get-UserProfiles.

BTW for 4.0 I would like to remediate the fact that some of these excludes are switches and some booleans, and adopt a similar approach to Invoke-HKCURegistrySettingsForAllUsers.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -5567,6 +5567,22 @@ Use Robocopy to copy files rather than native PowerShell method. Robocopy overco
 
 Additional parameters to pass to Robocopy. Default is: $null
 
+.PARAMETER ExcludeNTAccount
+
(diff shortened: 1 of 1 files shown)


mmashwani (approved):
Looks good, agree with the changes.

PR #956, opened 2024-04-16
Load 'AppDeployToolkitExtensions.ps1' after the $installName ist conf…

Basti890:
…igured

Currently you can't use the varibale $installname in the toolkit extension to define other custom varibales as this variable is set AFTER the extension is being procesesd.


mjr4077au:
@Basti890 thank you, this has been merged in manually.

PR #983, opened 2024-05-01
Allow Convert-RegistryPath to work with key paths originating from re…

Mrman1093:
…gistry PSObjects.

I found that when using the registry functions included with PSADT,  Convert-RegistryPath prevents key paths originating from registry PSObjects from working. eg:

```
$myKey = Get-RegistryKey -Key "HKLM:\SYSTEM\CurrentControlSet\Control\COM Name Arbiter
Set-RegistryKey -Key $myKey.PSPath -Name 'ComDB' -Value $comDB
```
This PR resolves the issue by removing 'Microsoft.PowerShell.Core\Registry::' and 'Microsoft.PowerShell.Core\Registry::\\' from key paths that include them.


DanGough:
Seems like a good addition, I will review this later today.

DanGough:
I think it would be better to tackle this issue differently. The whole purpose of Convert-RegistryPath is to convert common text strings like HKLM\ into full PSPaths.

You want to enter an already fully qualified PSPath, but are just trimming things off only for them to be put back on.

If the 2nd half of the function was altered to aim for outputting the fully qualified 'Microsoft.PowerShell.Core\Registry::' instead of 'Registry::', then this would achieve the same outcome in a more elegant way I think?

Mrman1093:
Looking at this a bit closer today, that would work but it would require modifying the lines below it where it verifies that the root key is valid and swaps out HKCC for HKCU. It would also prevent the processing of $Wow6432Node which might not be apparent to the end user. If ensuring that $Wow6432Node is processed is a priority, things can be moved around to make sure that it is always processed and not just when the input string doesn't start with 'Registry::'

Issue #999, opened 2024-05-23
[Bug] Activesetup not applying for current logged on user


FRacine:
### Prerequisites

- [X] Ensure you write a short, descriptive title after [Bug] above.
- [X] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [X] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

3.10.1

### Describe the bug

This was working previously:
wscript.exe "C:\Users\Public\PSAppDeployToolkit\PSAppDeployToolkit-ExecuteAsUser.vbs"

Our active setup file is there:C:\Program Files\Organisation\InstTrousses\Trousses\WinSCP6_Frv1\ActiveSetup\ActiveSetup.ps1
Manually it is working.
Running the whole product installation under a loggedon user with a runas admin  is working correctly
Under the system account it is not working.

### Steps to reproduce

1.Open an admnisitrator command window
2. Using psexec, run psexec -s -i cmd.exe
3.Start powershell
4. Run the script
```
5. **Set-ActiveSetup -StubExePath "$var_ActiveSetup83" -Description "Configuration de $var_PackageLongName" -Key "$var_PackageLongName" -Locale 'Fr'**

$var_ActiveSetup83
C:\PROGRA~1\ORGANI~1\INSTTR~1\Trousses\WINSCP~1\ACTIVE~1\ACTIVE~1.PS1

"Configuration de $var_PackageLongName"
Configuration de WinSCP6_Frv1

"$var_PackageLongName" 
WinSCP6_Frv1

7. =====================
**From function: Set-ActiveSetup**
Execute-ProcessAsUser -Path $CUStubExePath -Parameters $CUArguments -Wait -ContinueOnError $true

$CUStubExePath
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe

$CUArguments
-ExecutionPolicy Bypass -NoProfile -NoLogo -WindowStyle Hidden -Command "& {& \"C:\PROGRA~1\ORGANI~1\INSTTR~1\Trousses\WINSCP~1\ACTIVE~1\ACTIVE~1.PS1\"}"
=====================

8. =====================
**Function Execute-ProcessAsUser**
[...]

PR #1017, opened 2024-07-07
Set-ActiveSetup / Execute-ProcessAsUser fixes

DanGough:
# Pull Request

## Description

Fixes: #999 

### Bugs fixed:
- When Set-ActiveSetup was called with the StubPath being a direct path to a .ps1 file, it would fail to execute for the current user.
- Execute-ProcessAsUser incorrectly processed arguments when the main exe being called was cmd.exe; double quotes " were being replaced by ^".

### Enhancements:
- Set-ActiveSetup now injects wscript.exe into the StubPath instead of cscript.exe when passed a .vbs file. It makes sense to do so seeing how it goes to great lengths to ensure that any any .ps1 / .cmd files are run hidden.
- Set-ActiveSetup now supports .bat files (handled in the same way as .cmd files)

## Type of change

## Checklist

## How Has This Been Tested?

Create the following test scripts under C:\Temp:

**Say Hello.ps1:**
```powershell
Add-Type -AssemblyName System.Speech; $speak = New-Object System.Speech.Synthesis.SpeechSynthesizer; $speak.Speak('Hello')
```

**Say Hello.cmd: / Say Hello.bat:**
```vbscript
PowerShell -Command "Add-Type -AssemblyName System.Speech; $speak = New-Object System.Speech.Synthesis.SpeechSynthesizer; $speak.Speak('Hello')"
```

**Say Hello.vbs:**
```batch
Set sapi = CreateObject("SAPI.SpVoice")
sapi.Speak "Hello"
```

Then running any of the commands below should execute the script hidden, producing speech:

```powershell
Set-ActiveSetup -Key 'Test' -PurgeActiveSetupKey; Set-ActiveSetup -Key 'Test' -StubExePath 'C:\Temp\Say Hello.ps1'
Set-ActiveSetup -Key 'Test' -PurgeActiveSetupKey; Set-ActiveSetup -Key 'Test' -StubExePath 'C:\Temp\Say Hello.cmd'
Set-ActiveSetup -Key 'Test' -PurgeActiveSetupKey; Set-ActiveSetup -Key 'Test' -StubExePath 'C:\Temp\Say Hello.bat'
Set-ActiveSetup -Key 'Test' -PurgeActiveSetupKey; Set-ActiveSetup -Key 'Test' -StubExePath 'C:\Temp\Say Hello.vbs'
```


DanGough:
@mmashwani I'd appreciate your review here particularly around the change to RunHidden.vbs!

DanGough:
> We're changing all the commits on our PSADT Files, but it's not working. It only works if the path to activesetup.cmd has no spaces. We believe this is an issue of quotes being passed from the powershell script to the vbs script.

This scenario was tested in the example I gave above:

```powershell
Set-ActiveSetup -Key 'Test' -PurgeActiveSetupKey; Set-ActiveSetup -Key 'Test' -StubExePath 'C:\Temp\Say Hello.cmd'
```

Are you using similar syntax or something different?

Also please ensure you are using this branch:

https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/tree/DanGough/issue999

In Zip form:

https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/archive/refs/heads/DanGough/issue999.zip

DanGough:
There are now Pester tests for various use cases of Execute-ProcessAsUser and Set-ActiveSetup.

- RunHidden.vbs has been simplified. There is no need to escape metacharacters in here, it was just breaking the commands. Escaping of metacharacters with cmd.exe should be left to the user, EXCEPT when they just supply a .cmd / .bat file that has one of these characters in the name. RegEx replace now does this for such filenames.
- To stop vbscript dropping the quotes from the supplied args, they are swapped for a placeholder [{quote}] which is then turned back to " inside the .vbs.
- Quoting of cmd.exe args changed to ensure they are in the format cmd.exe /c ""path to file.cmd" arg1 "arg 2" arg3"

sintaxasn (approved):
Great work on this!

mmashwani (approved):
Nice, looks good!

DanGough:
Small extra code cleanup + a couple more test cases, all tests still passing.

IREJDW:
Hi 


Git Stubpath
if brackets occur in a variable such as env:Programfiles(x86), ^ characters are generated in the registry in the StubPath Value. See screenshots.

IREJDW:
you can test it with:
$StubExePath= "C:\()"
StubExePath = $StubExePath -replace '([()%!^&])', '^$1

Output of $StubExePath:
C:\^(^)

DanGough:
This was done When passing certain characters on the command line to cmd.exe, the characters ()%!^& have to be pre-fixed with ^ in order to work. However, I believe you do not need to do this if they are between double quotes, so I will take another look at this.

e.g. the brackets here would need to be escaped:

`cmd.exe /c IF EXIST file.txt ^(do something^)`

But these would not:

`cmd.exe /c del "C:\Program Files (x86)\file.txt"`

DanGough:
> Hi
> 
> Git Stubpath if brackets occur in a variable such as env:Programfiles(x86), ^ characters are generated in the registry in the StubPath Value. See screenshots. However, it works with env:Programfiles.

It turns out that brackets need to be escaped only if there is no space in the path. If there is a space, cmd.exe handles it differently and does not require brackets to be escaped, hence why `Program Files (x86)` works just fine.

You may see that `C:\Temp\Test(1).cmd` would need to be escaped to `C:\Temp\Test^(1^).cmd`, yet `C:\Temp\Test (1).cmd` would not!

One small further change pushed to Develop branch to correct this:

https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/b2351b0d64f18686d699a0f111be15cd44a5a60f

IREJDW:
Hello everyone,

the last adjustment seems to work now. I will test it but will not be able to do so for the next few days.

Thanks for the fix

Issue #1029, opened 2024-07-25
Async operations (e.g. Show-InstallationPrompt) triggers rotation of a new log file when <Toolkit_LogAppend> is set to False


DanGough:
### Prerequisites

- [X] Ensure you write a short, descriptive title after [Bug] above.
- [X] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [X] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

3.10.1

### Describe the bug

When AppDeployToolkitConfig.xml is configured with:

```xml
<Toolkit_LogAppend>False</Toolkit_LogAppend>
```

Then the code below will trigger the generation of a new log file purely for the prompt below:

```powershell
Show-InstallationPrompt -Message 'You can customize text to appear at the end of an install or remove it completely for unattended installations.' -ButtonRightText 'OK' -Icon Information -NoWait
```

### Steps to reproduce

See description.

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 23H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250
OsLanguage           : en-GB
OsMuiLanguages       : {en-GB, en-US}
KeyboardLayout       : en-GB
TimeZone             : (UTC+00:00) Dublin, Edinburgh, Lisbon, London
HyperVisorPresent    : True
CsPartOfDomain       : False
CsPCSystemType       : Mobile

.NET SDK:
 Version:           8.0.206
 Commit:            bb12410699
 Workload version:  8.0.200-manifests.f6879a9a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22631
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.206\

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.7
  Architecture: x64
  Commit:       2aade6beb0

.NET SDKs installed:
  8.0.206 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
[...]

Issue #1024, opened 2024-07-18
[Bug] Logs created during user context execution are not allowed to be altered by other users


Kevinnns:
### Prerequisites

- [X] Ensure you write a short, descriptive title after [Bug] above.
- [X] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [X] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

3.10.1

### Describe the bug

When you run PSADK in the user context it logs to c:\programdata\logs\software\*.log per default.

However the way permissions are granted withing c:\programdata\ allows users to create new files, but they are not allowed to alter files another user created.

So in the case of PSADK user A runs the script first, that user becomes owner of the file.
User B logs on and then runs the same PSADK script and will get errors as that user is not allowed to append the previously created log file.

### Steps to reproduce

1. In AppDeployToolkitConfig.xml set <Toolkit_RequireAdmin> to False
2. Execute PSADK script as user A (user should not have admin permissions) -> log file will be created in c:\programdata\logs\software
3. Execute PSADK script as user B (user should not have admin permissions) -> no logs will be added due to permission issues

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 23H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250
OsLanguage           : en-US
OsMuiLanguages       : {en-US}
KeyboardLayout       : 00020409
TimeZone             : (UTC+01:00) Brussels, Copenhagen, Madrid, Paris
HyperVisorPresent    : True
CsPartOfDomain       : False
CsPCSystemType       : Mobile
```

Kevinnns:
Discussing with Dan Gough on Discord multiple options possible.

1. Do a change in the XML the include $env:username in the path defined <Toolkit_LogPathNoAdminRights>
[...]

PR #1030, opened 2024-07-25
Bug fixes for logging (#1024, #1029)

DanGough:
# Pull Request

## Description

- Fix for #1024 where users are unable to modify log files created by other users when not running elevated. Now the username is appended to the log file name if $IsAdmin is false.

- Fix for #1029 where async operations would trigger log file rotation when <Toolkit_LogAppend> was set to false. Now $LogFileInitialized is set to true if $AsyncToolkitLaunch is true $ReferredLogname is present.

## Type of change

## Checklist

## How Has This Been Tested?

Toolkot prepared with <Toolkit_RequireAdmin> and<Toolkit_LogAppend> set to False, with a script that installs an MSI and calls Show-InstallationPrompt with -NoWait. Run Deploy-Application.exe twice each with and without elevation:

image

image

Previously in 3.10.1, there would be no _admin appended to the log files under ProgramData, and there would be 4 log files in each folder since Show-InstallationPrompt would generate another one.


mjr4077au:
It's not a big deal, but you could just do `[Boolean]$LogFileInitialized = $AsyncToolkitLaunch -and $ReferredLogname`, avoiding the unnecessary branching and achieves the same result.

DanGough:
Thanks @mjr4077au , actually I just spotted an area in the code where it relaunches the toolkit with -AsyncToolkitLaunch but without -ReferredLogname, so just going to do some further tests there.

sintaxasn (approved):

Issue #1031, opened 2024-07-26
Test-ServiceExists failes on Sandbox - Get-WmiObject : Provider load failure 


AlkHacNar:
### Prerequisites

- [X] Ensure you write a short, descriptive title after [Bug] above.
- [X] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [X] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

3.10.1

### Describe the bug

I tryed to get the Show-InstallationWelcome command to work with the serviceui. on a test notebook it worked fine, but it failed in the Windows Sandbox, cause it couldn't check for the service
`[Initialization] :: Failed check to see if service [Schedule] exists.`
I found out, that the Test-ServiceExists didn't work properly, first because of the `$env:ComputerName`, which is no promlem to fix int he sandbox with `$(hostname)`. But then it still didn't run with the failure `Get-WmiObject : Provider load failure `.
I don't know if there are more functions, which aren't working as expected in the sandbox and I know that psadt isn't really build to run in the sandbox, but I'd just wanted you to know. Maybe there is a fix/ workaround for the sandbox for this?

### Steps to reproduce

1.Show-InstallationWelcome
2.Test-ServiceExists

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 22H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250
OsLanguage           : de-DE
OsMuiLanguages       : {de-DE, cs-CZ, en-GB, en-US...}
KeyboardLayout       : en-US
TimeZone             : (UTC+01:00) Amsterdam, Berlin, Bern, Rome, Stockholm, Vienna
HyperVisorPresent    : True
CsPartOfDomain       : False
CsPCSystemType       : Mobile
```

mjr4077au:
This has already been addressed in v4.0 but can't be addressed here as we need to use the legacy WMI cmdlets for backwards compatibility.

AlkHacNar:
[...]

PR #1035, opened 2024-08-01
Allow service functions to work in Windows Sandbox

DanGough:
# Pull Request

## Description

- Service functions now use Get-Service instead of WMI where possible. Only exception is Test-ServiceExists will still use WMI if -PassThru supplied, so that it can return output in the same format as previously, for backwards compatibility. Windows Sandbox does not have the required WMI classes for services available.

- Service functions use 'localhost' as default computer name instead of $env:ComputerName. In Windows Sandbox, this name is truncated and is not the same as the DNS hostname, so fails.

Fixes: #1031

## Type of change

## Checklist

## How Has This Been Tested?

Tested functions work as before in a regular VM and also Windows Sandbox.

### DIFF
--- a/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
+++ b/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
@@ -1324,7 +1324,7 @@ https://psappdeploytoolkit.com
                     }
 
                     # Get all log files (including any .lo_ files that may have been created by previous toolkit versions) sorted by last write time
-                    $LogFiles = @(Get-ChildItem -LiteralPath $LogFileDirectory -Filter ("{0}_*{1}" -f $LogFileNameWithoutExtension, $LogFileExtension)) + @(Get-Item -LiteralPath ([IO.Path]::ChangeExtension($LogFilePath, 'lo_')) -ErrorAction Ignore) | Sort-Object LastWriteTime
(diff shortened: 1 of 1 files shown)


Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1
```diff
@@ -15362,29 +15362,25 @@ https://psappdeploytoolkit.com
...
+            If ($PassThru) {
+                $ServiceObject = Get-WmiObject -ComputerName $ComputerName -Class 'Win32_Service' -Filter "Name='$Name'" -ErrorAction 'Stop'
+                # If nothing is returned from Win32_Service, check Win32_BaseService
+                If (-not $ServiceObject) {
+                    $ServiceObject = Get-WmiObject -ComputerName $ComputerName -Class 'Win32_BaseService' -Filter "Name='$Name'" -ErrorAction 'Stop'
+                }
+            }
+            else {
+                # Known issue that WMI method fails in Sandbox, so only use it if -PassThru requested
+                $ServiceObject = (Get-Service -ComputerName $ComputerName -Name $Name -ErrorAction Ignore).Count -gt 0
```
mjr4077au: This can't use `-ErrorAction Ignore` is 3.x needs to remain compatible with PowerShell 2.x targets.

mjr4077au: The double bang notation `!!(Get-Service -ComputerName $ComputerName -Name $Name -ErrorAction SilentlyContinue)` is safer here as the automatic `Count` property isn't available on non-array object types when the StrictMode version is 2 or higher.

DanGough: TIL, I'll change that! It's used in a couple of other places that I resolved also.

Issue #1034, opened 2024-07-30
[Bug] Function Block-AppExecution not working if..


cyborgrd:
### Prerequisites

- [X] Ensure you write a short, descriptive title after [Bug] above.
- [X] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [X] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

3.10.1

### Describe the bug

Function Block-AppExecution not working if there is & in the [string]$installTitle like
[string]$installTitle = "Test & QA"

Importing of the XML fails 100% of 100%

![image](https://github.com/user-attachments/assets/3653c9ef-3b25-4811-943e-6e0e9fe8f4be)

### Steps to reproduce

Create Application with [string]$installTitle = "Test & QA"

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 23H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-Bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250
```

DanGough:
Looks like we may need to escape & in the XML via `[System.Security.SecurityElement]::Escape($string)` here.

DanGough:
Fixed in the PR above.

https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/pull/1036/commits/fe105f3eb72b32656bafa8600af789ce9a5d9968

PR #1036, opened 2024-08-01
Fix escaping of invalid characters in Block-AppExecution

DanGough:
# Pull Request

## Description

- Install name no longer goes through a list of illegal characters - only double quotes need to be substituted as they trip up the calling of VBScript. Since they are just in names/titles, they are replaced with two single quotes.

- $SchInstallName was a sanitized version of $InstallName, now there is $SchInstallTitle as a sanitized version of $InstallTitle.

- [System.Security.SecurityElement]::Escape used to escape entire command before insertion into the XML. This properly esacapes characters like & and ".

Fixes: #1034

## Type of change

## Checklist

## How Has This Been Tested?

Running a series of tests with Block-AppExecution with $InstallName and $InstallTitle set to contain all the characters that were previously being replaced - e.g:

```powershell
    [String]$installTitle = @'
Hi "there" $!'`();\*?{}[]<>|&%#~@'
'@
```

image


mjr4077au:
@DanGough can you please force push the PR so that the commits to be ingested are clean?

DanGough:
@mjr4077au done!

sintaxasn (approved):

PR #1038, opened 2024-08-07
Replace Select-Object -SkipLast with PowerShell 2 compatible equivalent

Mrman1093:
# Pull Request

## Description

Update `Copy-File` function to use a PowerShell 2 compatible alternative to `Select-Object -SkipLast 1` 

The `-SkipLast` parameter was added to `Select-Object` in PowerShell 5
https://learn.microsoft.com/en-us/previous-versions/powershell/module/microsoft.powershell.utility/select-object?view=powershell-4.0

https://learn.microsoft.com/en-us/previous-versions/powershell/module/microsoft.powershell.utility/select-object?view=powershell-5.0

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] I am pulling to the **develop** branch
- [X] I have made corresponding changes to the documentation

## How Has This Been Tested?

Dot-sourced AppDeployToolkitMain.ps1 and used `Copy-File` to copy a folder from one location to another.


DanGough:
Previous code always made the assumption that the last line was to be ditched. Your change will end up returning the single line twice if the output is a single line... I think a better approach is to simply trim the entire output to get rid of the last blank line, then use a regex replace to get rid of any whitespace at the beginning of each line:

```powershell
$RobocopyOutput = $RobocopyResult.StdOut.Trim() -Replace '\n\s+',"`n"
```

DanGough (approved):
PR updated with neater implementation. Good to go!

sintaxasn (approved):

PR #1075, opened 2024-11-25
Allow PSAppDeployToolkit v4.0 config options to be specified via the registry as administrative overrides

mjr4077au:
# Pull Request

## Description

In support of https://github.com/PSAppDeployToolkit/PSAppDeployToolkit4-devel/issues/55, configuration parsing now works in the following fashion:

* Module's default config is imported.
* Caller's config adjacent to their `Invoke-AppDeployToolkit.ps1` script is superimposed over the top.
* Now, registry values defined via Intune or Group Policy are superimposed over the top of those.

This will facilitate mass deployment changes such as dialog styles, etc, without necessarily having to repackage all applications.

Fixes: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit4-devel/issues/55

## Type of change

## Checklist

## How Has This Been Tested?

This has been tested with the relevant registry settings defined in HKEY_LOCAL_MACHINE\SOFTWARE\PSAppDeployToolkit\Config and works as expected.

image

Supporting ADMX files should be created to support the administrator and allow for natural management without resorting to scripts, etc.

### DIFF
--- a/src/PSAppDeployToolkit/Config/config.psd1
+++ b/src/PSAppDeployToolkit/Config/config.psd1
@@ -37,7 +37,7 @@
         # Specify if the log files should be bundled together in a compressed zip file.
         CompressLogs = $false
 
-        # Choose from either 'Native' for native PowerShell file copy via Copy-Item, or 'Robocopy' to use robocopy.exe.
+        # Choose from either 'Native' for native PowerShell file copy via Copy-ADTItem, or 'Robocopy' to use robocopy.exe.
         FileCopyMode = 'Native'
 
         # Specify if an existing log file should be appended to.
@@ -77,7 +77,7 @@
         # Same as RegPath but used when RequireAdmin is False. Bear in mind that since this Registry Key should be writable without admin permission, regular users can modify it also.
         RegPathNoAdminRights = 'HKCU:\SOFTWARE'
 
-        # Specify if Administrator Rights are required. NB: Some functions won't work if this is set to false, such as deferral, blockexecution, file & registry RW access and potentially logging.
+        # Specify if Administrator Rights are required. Note: Some functions won't work if this is set to false, such as deferral, block execution, file & registry RW access and potentially logging.
         RequireAdmin = $true
 
         # Automatically changes DeployMode for session zero (SYSTEM) operations.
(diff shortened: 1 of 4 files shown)


DanGough (commented):
Are we happy with the reg key being `HKLM\SOFTWARE\PSAppDeployToolkit`?

Would an ADMX template put entries under `HKLM\SOFTWARE\Policies\PSAppDeployToolkit` by default? We might also want to consider if a policy could be applied at a user or device level, in which case we might also want to look at the HKCU equivalent, and decide which one takes precedent.

Since I would refer to the settings themselves as Policies, and consider ADMX as just one way to set those keys, I would rename a couple of things:

`-NoAdmxParsing` -> `-NoPolicy` or `-IgnorePolicy` perhaps?

`$admxSettings` -> `$policySettings`

mjr4077au:
> Are we happy with the reg key being `HKLM\SOFTWARE\PSAppDeployToolkit`?

It could certainly move there, though initially I was wanting to keep all the keys pertaining to PSAppDeployToolkit contained to one location. Not all ADMX templates manipulate keys within a Policies subkey, they can technically be wherever you want them to be within the registry.

> Would an ADMX template put entries under `HKLM\SOFTWARE\Policies\PSAppDeployToolkit` by default? We might also want to consider if a policy could be applied at a user or device level, in which case we might also want to look at the HKCU equivalent, and decide which one takes precedent.

Whatever an admin defines within the registry, either manually or via an ADMX when we get to it would be the authoriative value used, superseding a local config along-side a deployment script, and of course the module's default config.

Regarding settings in HKCU, I don't think it's a good idea as users could technically stuff their own settings in there and manipulate an installation surruptitiously. I think machine-based policies only would be the best approach.

> Since I would refer to the settings themselves as Policies, and consider ADMX as just one way to set those keys, I would rename a couple of things:
> 
> `-NoAdmxParsing` -> `-NoPolicy` or `-IgnorePolicy` perhaps?
> 
> `$admxSettings` -> `$policySettings`

Yeah, I can change that out. I prefer `-IgnorePolicy` to `-NoPolicy` if you're amenable to that?

mjr4077au:
Hi @DanGough, I've changed the base registry path as suggested, and the `-NoAdmxParsing` switch to `-IgnorePolicy`. Please let me know if there's any further review items.

PR #1094, opened 2024-12-04
Unofficial PSADT module cleanup

mjr4077au:
# Pull Request

## Description

This PR adds the necessary logic to remove the unofficial PSADT module from systems where found, and handles situations whereby the caller may be in PowerShell 7 but the PSADT module was installed via Windows PowerShell 5.x, etc.

It should be noted that this PR does increase the module importation time, especially when the module _is_ present and requires removal. Typical import time for the module is ~3 seconds. Following this change, if PSADT isn't found, ~7 seconds. Following this change and PSADT is found, ~10 seconds.

The penalty of import duration is worth considering as the unofficial PSADT module has been a painful experience and often collides with users and their intention as they don't understand the clobbering aspect of having that module loaded/available while attempting to use PSADT 3.x independently of it.

## Type of change

## Checklist

## How Has This Been Tested?

This has been tested on my device end to end in Windows PowerShell 5.x and PowerShell 7.

### DIFF
--- a/src/PSAppDeployToolkit/ImportsFirst.ps1
+++ b/src/PSAppDeployToolkit/ImportsFirst.ps1
@@ -56,22 +56,16 @@ $CommandTable = [ordered]@{}; $ExecutionContext.SessionState.InvokeCommand.GetCm
 # Ensure module operates under the strictest of conditions.
 & $CommandTable.'Set-StrictMode' -Version 3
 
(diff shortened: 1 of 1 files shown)


sintaxasn (approved):
Downgradede to a Warning and removed the dependency on PowerShellGet. Good to go.

PR #1102, opened 2024-12-05
PSAppDeployToolkit 4.0.2

mjr4077au:
* Patch out null access for `$envLogonServer` within `New-ADTEnvironmentTable` which is preventing proper use with SCCM.
* Fix bad setup in `Show-ADTBalloonTip` that was preventing `Get-ADTBoundParametersAndDefaultValues` from working right.
* Revise `Get-ADTBoundParametersAndDefaultValues` to exclude common parameters by default.


sintaxasn (approved):
Good to go

PR #1110, opened 2024-12-05
Partial re-write of `Start-ADTMsiProcess`.

mjr4077au:
* Return earlier when attempting to install an already-installed MSI, or performing any other action against a non-installed MSI.
* Ensure that ProductCode throws if used with an Install action.
* Improve setup for `$msiFile` (now renamed to `$msiProduct`).
* Rework log filename setup within function to better prevent null accesses. Fixes #1105n
* Rework log filename setup for consistency when deriving a filename from an InstalledApplication object vs an .msi FilePath.
* Rearrange where transformations and patches have their paths fully resolved since transforms are sent through to `Get-ADTMsiTableProperty`.
* Rename a few internal variables for clarity.


DanGough (approved):
All good, tested various installs/uninstalls/repairs and working as expected.

PR #1113, opened 2024-12-05
Update strings.psd1

ClientMGMT:
Fixed a couple of translation errors for the WelcomePrompt message text in Swedish (sv).

Fixed the text for DialogMessageNoProcesses to make sense in swedish and fixed the ButtonDeferRemaining text so it uses the correct word.


mjr4077au:
@ClientMGMT thank you very much!

PR #1207, opened 2024-12-19
PSAppDeployToolkit 4.0.4

mjr4077au:
- Ensure the build system removes `Invoke-AppDeployToolkit.pdb` from the v3 frontend folder.
- Get custom scripts working with the exe as the parameter parsing was wrong.
- Ensure a message dialog from the exe is always topmost.
- Reset `PSModulePath` environment variable in the ext to handle pwsh.exe > exe > powershell.exe invocations.
- Fix config/strings path print issues within `DeploymentSession`'s constructor.
- Ensure Show-ADTDialogBox is bypassed in Non-Interactive mode, because it's interactive
- Make all string comparisons for arguments case insensitive
- Ensure PSADT.Invoke doesn't show any messages if running in NonInteractive/Silent modes.
- Throw a better exception on PowerShell process failure.
- Catch `MessageBox.Show()` exceptions in PSADT.Invoke and swallow them.
- Amend check in module initialisation code to prevent localisation issues.
- Extend previous commit to make `Open-ADTSession`'s `-ScriptDirectory` parameter a string array also.
- Allow `Initialize-ADTModule` to accept an array of ScriptDirectory values.
- Synchronise changes within `Invoke-AppDeployToolkit.ps1` with the example scripts.
- Revert most of "- Don't rely on `$PSBoundParameters` within `Invoke-AppDeployToolkit.ps1` so callers are free to extend the parameters with their own variables if desirable." as it's no longer required.
- Rework the architecture setup to always use the x64 ServiceUI executable on any 64-bit target.
- Rework arguments setup as you can't pass booleans to SwitchParameter values via `powershell.exe -File <script.ps1>` setups.
- Ensure `Invoke-ServiceUI.ps1` stops on a dime if there's an error.
- Ensure ServiceUI wrapper batch scripts exit with PowerShell's exit code.
- Fix bad `$StartMode` re-writing within `Start-ADTServiceStartMode`.
- Change last remaining `Write-Error` call to a throw on the back of the previous commit.
- Relax the duplicate assembly checking stringency by allowing existing assemblies if the hashes match the currently importing module's assemblies.
- Rework module importing to remove remaining `$RequiredModules` from module's global scope.
- Relocate a throw in `ImportsFirst.ps1` so that it doesn't break due to $CommandTable substitution which doesn't exist yet.
- Optimise function imports to prevent cycling through function names twice.
- Extend previous commit to the unbuilt PSAppDeployToolkit.psm1 file as well.
- Make the build system do $CommandTable translations in `ImportsFirst.ps1`/`ImportsLast.ps1`.
- Use PowerShell's `Add-Type` for the normal assembly load pathway.
- Optimise previous commit by only doing network checks once, and ensure changed code uses our $CommandTable.
- Add check for mapped drive
- Import our assemblies via the psm1 file so we can load them from UNC paths if required.
- Rework module import via manifest check to not be dependent on the assembly.
- Add a bit more modularity to the build system.
- Resolve `New-ADTTemplate`'s `-Destination` parameter prior to using it.
- Set the DefaultUser SID in `Get-ADTUserProfiles` using `WellKnownSidType.NullSid` instead of a string literal.
- When compiling configs/string tables, don't erase defined values with null/empties from successive imports.
- Ensure `-Force` is honoured in `Exit-ADTInvocation`.
- Make `Close-ADTSession` throw if it's called without an active session.
- Remove leftover markdown help for `Show-ADTBlockedAppDialog`.
- Set up proper enums for `DeploymentType` and `DeployMode` as well.
- Fix module build system following IFEO changes/fixes.
- Minimise amount of `Get-ADTStringTable` calls within `Block-ADTAppExecution` and add some quote safety.
- Amend previous commit to use `conhost.exe --headless` to hide the window.
- Fix IFEO path length constraints potentially effecting BlockExecution functionality.
- Fix recovery scheduled task creation in `Block-ADTAppExecution`.
- Fix bad SwitchParameter evaluation when constructing a `DeploymentSession` object.
- Sub-express `Get-ADTRegistryKey` calls in `Get-ADTUserProfiles` to avoid typing errors.
- Optimise previous commit to ensure a session is always removed from the buffer during a close failure.
- Ensure a failed `Close-ADTSession` operation exits out with an exit code.
- Update markdown help export for `Open-ADTSession`.
- Pack `DeploymentSession`'s numerous bools into a bitfield.
- Change `DeploymentSession.GetDeploymentStatus()` to be a proper enum.
- Don't flush the module's callbacks upon module init otherwise it defeats the purpose of the system.
- Allow specifying a `DeploymentSession` class override against `Open-ADTSession`.
- Add better throws to `InternalDatabase`'s getters.
- Store out the SessionState within `InternalDatabase` to prevent repeated lookups.
- Ensure all `DeploymentSession` scriptblock invocations occur via the module's SessionState.
- Change loading of `PSAppDeployToolkit.Extensions` to wildcard search for `PSAppDeployToolkit.*`.
- Rework process closing code when entering the `-PromptToSave` branch.
- Provide more/better filtration within `Get-ADTWindowTitle`.
- Improve the output description in `Get-ADTApplication`'s comment-based help.
- Flip a branch around in `Show-ADTInstallationWelcome` as it's cleaner.
- Fix bad setup in `DeploymentSession.Close()` stemming from bad auto-completion during "- Initial porting of ADTSession.Close() into C# code.".
- Finalise implementation of "- Rework DeploymentSession instantiation to not require sending in module internals during construction." by removing private database members from `DeploymentSession`.
- Move DeploymentSession list addition/removal to its constructor and close methods, respectively.
- Rework DeploymentSession instantiation to not require sending in module internals during construction.
- Rework `DeploymentSession`'s `RunspaceOrigin` parameter to be `NoExitOnClose`, which is what it actually means.
- Silence null warning for `_installTitle` within `DeploymentSession`'s ctor.
- Make the exe pass through `-NonInteractive` on the powershell.exe command line.
- Rework `Open-ADTSession`s `$runspaceOrigin` setup so that setups via functions don't exit powershell.exe unless `-NonInteractive` is explicitly passed.
- Fix bad `_installName` regex setup within `DeploymentSession`'s constructor.
- Add `Convert-ADTValueType` to provide casted value type conversions without overflow exceptions.
- Update markdown help export for `New-ADTTemplate`.
- Optimise the algorithm within `Set-ADTPreferenceVariables`.
- Mock `Set-ADTPreferenceVariables` within all Pester unit tests due to the expense of the algorithm.
- Rework how we manage caller preference values to make the implementation more robust between modules.
- Avoid using `$PWD` as it's not immutable.
- Unseal `PSADT.Module.DeploymentSession]` to allow inheritance for advanced customisations.
- Have our build system sign our own compiled wpfui DLL files.
- Integrate our local wpfui library into our PSADT.UserInterface project and rebuild our DLL files.
- Update all packages used by wpfui to the latest versions.
- Import wpfui, pegged at commit c3a50f74ee76f53cabea9f0d19e1287e56fb15dd (4.0.0-rc.2).
- Ensure throws from DeploymentSession's constructor maintain the exception's stack trace that it's rethrowing.
- Fix bad setup stemming from "- Add `DisableDefaultMsiProcessList` to `Open-ADTSession` to allow disabling the zero-config MSI process list evaluation.".
- Add `Get-ADTCommandTable` to the module.
- Change the module's CommandTable from an OrderedDictionary to a ReadOnlyDictionary.
- Improve exception handling within `PSADT.GUI.Explorer.RefreshDesktopAndEnvironmentVariables()`.
- Restore lost log message within `Update-ADTDesktop`.
- Optimise setup in `Get-ADTPEFileArchitecture` to avoid an unnecessary try/catch block.
- Remove `$runningProcess.Refresh()` line from `Show-ADTInstallationWelcome`.
- Add new function `Get-ADTOperatingSystemInfo`.
- Add `DisableDefaultMsiProcessList` to `Open-ADTSession` to allow disabling the zero-config MSI process list evaluation.
- Use `$PSCmdlet.ThrowTerminatingError()` within `Show-ADTWelcomePromptFluent` since it has CmdletBinding().
- Minor refactor for `Show-ADTWelcomePromptFluent` to optimise out an extra `Get-ADTRunningProcesses` call.
- Drop unused `Get-GitHubReleaseAssetUri` function from the build system.
- Add `Get-ADTPresentationSettingsEnabledUsers` to test whether the user has enabled presentation mode via the mobility settings on the device.
- Add stronger typing to `Get-ADTUserProfiles` returned object output.
- Add `-SkipUnloadedProfiles` parameter to `Invoke-ADTAllUsersRegistryAction`.
- Add in some empty files to the Files/SupportFiles folders to stop GitHub upload-artifact from dropping the empty folders.
- Fix incorrect comparison operator from "- Advise that `-WindowLocation` has no effect for `Show-ADTInstallationProgress` with fluent dialogs.".
- Add `New-ADTExampleFunction` as an example function within `PSAppDeployToolkit.Extensions`.
- `string.IsNullOrWhiteSpace()` always.
- Make `DynamicProcessEvaluation = $false` work right for `Show-ADTWelcomePromptFluent`.
- Allow overriding a session's default LogName value.
- Add config option to allow sending log files to a subfolder of LogPath.
- Remove all references to "XML" within comment-based help.
- Advise that `-WindowLocation` has no effect for `Show-ADTInstallationProgress` with fluent dialogs.
- Update OutputFile for build dlls to fix git commit time checks
- Write log timestamps in the same format 'git log' uses to aid troubleshooting
- Reduce verbosity of msbuild to minimal
- Add parentheses to path expressions used in git log commands, improve log messages
- Allow PDB files to be checked in to ease debugging with users remotely.
- Recompile checked in DLL file and remove guard clause from `Start-ADTProcessAsUser`.
- Allow the `CreateProcessAsUser()` code to work for non-SYSTEM admin users.
- Enable the SeDebugPrivilege for tokens created with the TokenManager class so that admins not running as SYSTEM can impersonate users as well.
- Update markdown help export for `Open-ADTSession`.
- Fix bad comment-based help within `Copy-ADTFile`.
- Numerous fixes to `Set-ADTActiveSetup`.
- Shoe-horn in process to allow custom variables to be added to a DeploymentSession object.
- Add a copy of the module's public certificate to the module root.
- Fix bad CIMMethod arguments setups from "- Replace bad `Get-CimInstance` calls with the correct `Invoke-CimMethod` calls.".
- Fix bad setup in `New-ADTShortcut`'s `-Path` ValidateScript decoration.
- Replace bad `Get-CimInstance` calls with the correct `Invoke-CimMethod` calls.
- Only use the .NET Framework 4.6.2 DLL files with our project.
- Fix badly escaped formatting in log compression string.
- Fix partial `platyPS` version bump from "- Bump platyPS version to 0.14.2, the latest released version.".
- Remove the `HelpInfoURI` value from the manifest, we don't support this.
- Updated wording of Norwegian.
- Bump platyPS version to 0.14.2, the latest released version.
- Update markdown exports following recent changes.
- Add `Export-ADTEnvironmentTableToSessionState` to toolkit to allow manual environment table exportation to the provided SessionState object.
- Restore `Get-ADTEnvironment` as a compatibility wrapper. To be removed at commencement of PSAppDeployToolkit 4.2.0.
- Rename `Get-ADTEnvironment` to `Get-ADTEnvironmentTable`.


sintaxasn (approved):
Amazing work!!

PR #1211, opened 2024-12-22
Add `-MaximumElapsedTime` parameter to `Invoke-ADTCommandWithRetries`

Mrman1093:
This parameter will allow you to retry a command until a given timespan has passed. If both a timespan and a retry limit have been provided it will continue until either one is met.

### DIFF
--- a/docs/Invoke-ADTCommandWithRetries.mdx
+++ b/docs/Invoke-ADTCommandWithRetries.mdx
@@ -28,7 +28,21 @@ This function invokes the specified cmdlet/function, accepting all of its parame
 Invoke-ADTCommandWithRetries -Command Invoke-WebRequest -Uri https://aka.ms/getwinget -OutFile "$($adtSession.DirSupportFiles)\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe.msixbundle"
 ```
 
-Downloads the latest WinGet installer to the SupportFiles directory.
+Downloads the latest WinGet installer to the SupportFiles directory. If the command fails, it will retry 3 times with 5 seconds between each attempt.
+
+### EXAMPLE 2
+```powershell
+Invoke-ADTCommandWithRetries -Command Get-FileHash -Path '\\MyShare\MyFile' -MaximumElapsedTime (New-TimeSpan -Seconds 90) -SleepSeconds 2
+```
+
+Gets the hash of a file on an SMB share. If the connection to the SMB share drops, it will retry the command every 2 seconds until it successfully gets the hash or 90 seconds have passed since the initial attempt.
+
+### EXAMPLE 3
+```powershell
+Invoke-ADTCommandWithRetries Copy-ADTFile -Path \\MyShare\MyFile -Destination C:\Windows\Temp -Retries 5 -MaximumElapsedTime (New-TimeSpan -Minutes 5)
+```
+
+Copies a file from an SMB share to C:\Windows\Temp. If the connection to the SMB share drops, it will retry the command once every 5 seconds until either 5 attempts have been made or 5 minutes have passed since the initial attempt.
 
 ## PARAMETERS
 
@@ -58,7 +72,7 @@ Parameter Sets: (All)
 Aliases:
 
 Required: False
-Position: 2
+Position: Named
 Default value: 3
 Accept pipeline input: False
 Accept wildcard characters: False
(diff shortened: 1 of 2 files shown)


docs/Invoke-ADTCommandWithRetries.mdx
```diff
@@ -28,7 +28,21 @@ This function invokes the specified cmdlet/function, accepting all of its parame
 Invoke-ADTCommandWithRetries -Command Invoke-WebRequest -Uri https://aka.ms/getwinget -OutFile "$($adtSession.DirSupportFiles)\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe.msixbundle"
 ```
 
-Downloads the latest WinGet installer to the SupportFiles directory.
+Downloads the latest WinGet installer to the SupportFiles directory. If the command fails, it will retry 3 times with 5 seconds between each attempt.
+
+### EXAMPLE 2
+```powershell
```
mjr4077au, line 34: Were these examples added by hand, or by our build system? They appear hand-added. If so, can you please:
1. Run `.\actions_bootstrap.ps1` from the root of our repository
2. Run `Invoke-Build -File .\src\PSAppDeployToolkit.build.ps1` to build the module, generating help files appropriately.

src/PSAppDeployToolkit/Public/Invoke-ADTCommandWithRetries.ps1
```diff
@@ -70,6 +87,10 @@ function Invoke-ADTCommandWithRetries
         [ValidateRange(1, 60)]
         [System.UInt32]$SleepSeconds = 5,
 
+        [Parameter(Mandatory = $false)]
+        [ValidateScript({ $_ -gt [System.TimeSpan]::New(0) })]
```
mjr4077au, line 91: Better to use `[System.TimeSpan]::Zero` here.

src/PSAppDeployToolkit/Public/Invoke-ADTCommandWithRetries.ps1
```diff
@@ -70,6 +87,10 @@ function Invoke-ADTCommandWithRetries
         [ValidateRange(1, 60)]
         [System.UInt32]$SleepSeconds = 5,
 
+        [Parameter(Mandatory = $false)]
```
mjr4077au, line 90: Can you put this into a separate parameter set? That'd be the cleaner way than allowing all parameters, but having one parameter override another with `$PSBoundParameter.ContainsKey()` checks.

src/PSAppDeployToolkit/Public/Invoke-ADTCommandWithRetries.ps1
```diff
@@ -106,17 +127,35 @@ function Invoke-ADTCommandWithRetries
...
                 {
+                    $i++
                     try
                     {
                         return (& $commandObj @boundParams)
                     }
                     catch
                     {
+                        $errorRecord = $_
+                        if ($PSBoundParameters.ContainsKey('MaximumElapsedTime') -and ($stopwatch.ElapsedMilliseconds -ge $MaximumElapsedTime.TotalMilliseconds))
```
mjr4077au, line 142: Rework this when it's in a different parameter set.

mjr4077au (changes requested):
Please have a look at the review comments.

src/PSAppDeployToolkit/Public/Invoke-ADTCommandWithRetries.ps1
Mrman1093: It's not meant to completely override the `-Retries` parameter, just when that parameter isn't provided.

docs/Invoke-ADTCommandWithRetries.mdx
Mrman1093: They were added by hand. I tried running the build process but ran into a bunch of errors. I'll try to get it to run again in the morning.

mjr4077au:
I've merged this in and applied some fixes over the top, mostly:
* https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/356a33ebf25ad89b0c4d1a30f84357d65c1413d3: To provide a better output to our callers, we throw errors in `[ValidateScript()]` decorations using `$PSCmdlet.ThrowTerminatingError()` with our own ErrorRecord out of `New-ADTValidateScriptErrorRecord`.
* https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/ea7abd275bb6ca45b7a156f987d7af1cfc026333: When using a proper stopwatch, we don't want to lose any of its precision so rather than comparing elapsed milliseconds, we can directly compare its Elapsed TimeSpan object with the value provided by the caller.
* https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/1cbc4d21bce98a3c9f2062eabd88de1c6d21c668: Empty control statements aren't great so I refactored this to avoid it. We also only instantiate a Stopwatch if the parameter has been specified, then we check whether the Stopwatch value is null to avoid querying the `$PSBoundParameters` dictionary over and over again.

I also took the opportunity to do https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/50e109a9d928974b95ebe3c59b36ee50231dc701, which is effectively getting this parameter into the correct state that it should always have been in.

PR #1320, opened 2025-02-09
- Add graceful stop process attempt for `close apps`

tjgruber:
Enhanced process termination logic to **attempt a graceful shutdown** before forcefully terminating processes. Instead of immediately calling `Stop-Process`, this update first calls `CloseMainWindow()` and waits up to **2 seconds** for the process to exit. If the process does not close within the timeout, it is then **forcefully terminated**.

This **reduces ungraceful terminations**, improving stability and minimizing unexpected prompts/errors in some applications.

#### **Scope & Intent**
This is an **incremental improvement**, not a full solution. Certain scenarios (e.g., multiple Chrome windows) may still require force termination.

#### **Test Case: Google Chrome**
- **Before**: Chrome prompts that it was not shut down properly and asks if tabs should be restored.
- **After**: Chrome reopens all tabs cleanly without the improper shutdown prompt.
- **Known Limitation**: If multiple Chrome windows are open, one may still be forcefully terminated.

This **should not impact other applications** but ensures that where possible, apps are given a chance to exit cleanly.

#### **Example Usage:**
```powershell
Show-ADTInstallationWelcome -CloseProcesses 'chrome' -Subtitle 'Company IT - Application Update' -AllowDeferCloseProcesses -CheckDiskSpace -ForceCountdown 10 -NoMinimizeWindows -DeferDays 5 -DeferRunInterval ([System.TimeSpan]::FromHours(24))
```

#### **Behavioral Flow with Above Use Case:**
1. If an application is running, the user is prompted with `Close Programs`, `Defer`, or `Continue`.
   - If `Close Programs` is selected, apps **gracefully close** (if supported).
   - If apps are manually closed before selecting `Continue`, they close gracefully.
   - If the **countdown timer expires**, `Defer` is used (if available), otherwise `Close Programs` is forced.
   - If the **deadline is reached**, `Defer` is **not available**.
   - If `Defer` is chosen or forced, the install/upgrade is retried in **24 hours**.

### DIFF
--- a/src/PSAppDeployToolkit/Public/Show-ADTInstallationWelcome.ps1
+++ b/src/PSAppDeployToolkit/Public/Show-ADTInstallationWelcome.ps1
@@ -947,8 +947,47 @@ function Show-ADTInstallationWelcome
                                 }
                                 else
                                 {
-                                    Write-ADTLogEntry -Message "Stopping process $($runningApp.Process.ProcessName)..."
-                                    Stop-Process -Name $runningApp.Process.ProcessName -Force -ErrorAction Ignore
+                                    try
+                                    {
+                                        $runningAppProcessName = $runningApp.Process.ProcessName
+
+                                        # Close out the main window
+                                        if ($runningApp.Process.CloseMainWindow())
+                                        {
+                                            $closeTimeout = [System.Diagnostics.Stopwatch]::StartNew()
+                                            $maxWaitTime = [System.TimeSpan]::FromSeconds(2)
+
+                                            do
+                                            {
+                                                [System.Threading.Thread]::Sleep(500)
+                                                $runningApp.Process.Refresh()
+                                            }
+                                            while (!$runningApp.Process.HasExited -and $closeTimeout.Elapsed -lt $maxWaitTime)
+
+                                            if ($runningApp.Process.HasExited)
+                                            {
+                                                Write-ADTLogEntry -Message "Process $runningAppProcessName closed gracefully."
+                                            }
+                                            else
+                                            {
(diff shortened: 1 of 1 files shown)


mjr4077au:
The problem here is that this isn't really addressing the underlying problem of why `-PromptToSave` isn't working, according to @DanGough and yourself from Discord. Are you going to be looking into that area at all? I swear that part of the code was firing though.

DanGough:
It would need some thorough testing, first to confirm there hasn't been a regression from V3 to current V4, then to see what this improves. I should be able to spend some time on it this week.

The help docs state that user interaction (e.g. psexec /s /i, or ServiceUI, or the native ConfigMgr option) is required for the -PromptToSave functionality to work... In future we may need to test whether closing processes running as the user should be closed via the thread running as the user.

tjgruber:
> The problem here is that this isn't really addressing the underlying problem of why `-PromptToSave` isn't working, according to @DanGough and yourself from Discord. Are you going to be looking into that area at all? I swear that part of the code was firing though.

Yes since the Discord conversation, I did find that it fires, and does attempt a graceful process closure. However, I think I misunderstood the intent of `-PromptToSave` due to the app I was testing with didn't behave as that code intends. 

It's clear to me that it's supposed to attempt to gracefully close a process that supports `CloseMainWindow()`, which it does. All is well there. It does work with using serviceui, with the exception that `BringWindowToFront` had no effect at least with the apps I was testing (which I don't care about). However it might work with something like Word.

I'll go ahead and close this out then since `-PromptToSave` is actually working mostly as intended, but can get weird depending on the app... which the same kind of issue is there even with this PR. However, this PR is a little more aggressive which was my preference, but is fine as-is.

PR #1323, opened 2025-02-10
Update strings.psd1 (NL)

CSVDK:
Improved overal text and consistancy of word use.


mjr4077au:
Thanks for the improvements!

PR #1433, opened 2025-04-28
Update Add-ADTModuleCallback.ps1

That-Annoying-Guy:
List hookpoints for Add-ADTModuleCallback


mjr4077au:
I'm not sure the values need explicitely calling out as the type is an enum, so the markdown help export will list the accepted values. https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/develop/docs/Add-ADTModuleCallback.mdx#-hookpoint

PR #1460, opened 2025-05-16
Updates to INI file functions

DanGough:
## Status

There was a recent change in the dev branch to rename and deprecate these functions:

- Get-ADTIniValue -> Get-ADTIniSectionKeyValue
- Set-ADTIniValue -> Set-ADTIniSectionKeyValue

And new functions added:

- Remove-ADTIniSectionKeyValue
- Remove-ADTIniSectionKey

## Context

Since there is no actual change in input/output/behaviour of these functions since v4.0 (at least this is the case after these changes), it seems this would cause unnecessary disruption.

As for the Remove function, we had 2 functions doing almost the same thing. A single Remove-ADTIniValue function could handle both situations and more closely mirror how functions like Remove-ADTRegistryKey behave, that can handle removal of both entire keys or specific values.

## Decision

After some discussion with @mjr4077au:
- The Get/Set function name changes have been reverted to maintain continuity.
- The Remove functions have been consolidated into one.

## Consequences

This change maintains continuity and simplifies the usage of the INI functions.

## Notes

In 4.0.6, the docs stated that sending a null value would erase the key, but due to a bug it actually set an empty value. I actually prefer that a Set function does this and we have a separate Remove function for explicitly removing an INI entry.

In this new implementation, if a null value is sent to Set-ADTIniValue, it will be converted to an empty string before being sent to Kernel32.WritePrivateProfileString so that the key is set to an actual empty value instead of being erased. This maintains the same behaviour as previous v4.0.x versions.

To enable **Set-ADTIniValue** to be able to set an empty value, I had to tweak **PInvoke.WritePrivateProfileString** to stop sending a null if **lpString** was NullorWhiteSpace. It now just checks directly if it's null.

This in turn broke **Remove-ADTIniValue**, as PowerShell was actually sending an empty string to the C# function instead of a null here:
`[PSADT.Utilities.IniUtilities]::WriteSectionKeyValue($Section, $Key, $null, $FilePath)`

This was fixed by sending `[NullString]::Value` instead:
`[PSADT.Utilities.IniUtilities]::WriteSectionKeyValue($Section, $Key, [NullString]::Value, $FilePath)`


mjr4077au:
All of the `[NullString]::Value` stuff is completely unnecessary as the type in C# is nullable, so sending null through like I had it was enough. Did you test the setup that was there before modifying it?

mjr4077au:
Apologies, I read your notes better and understand why the change was made. Still though, the null should have been enough there as we send in null values to other string constructor parameters elsewhere and those isn't an issue. I would have to testing here 

Regarding the renaming, I didn't do it on a whim. The 4.1.0 release is meant to correct a lot of things that weren't right with 4.0.0. The rename included the necessary compatibility wrappers for the release until people got to 4.2.0

Changing this back isn't necessarily the best move, and doing on on the basis of how `Remove-ADTRegistryKey` does it isn't right either because it implies the behaviour of that function is how it should be, which it isn't. 

Referencing PoweShell directly, you have `Remove-Item` and `Remove-ItemProperty'. Two functions, clearly named, that do exactly what the name says. This is precisely why there was `Remove-ADTIniSectionKey` and `Remove-ADTIniSectionKeyValue` as it mirrors that entirely and it's incredibly clear what you're doing based on the function name alone.

Further to this, having functions like `Get-ADTIniSectionKeyValue` which is clear in what it does opens us up to future expansion for things like `Get-ADTIniSection`, where we return a dictionary of all key/value pairs within a section. 

I know you said we discussed this, but we didn't discuss the wholesale reversion back to almost what was there before, and that's regrettable as you've probably spent the entire day on something I'm inclined not to merge. As such, I'm going to ask @sintaxasn for his review also so that argument can be 2:1 in a given direction.

mjr4077au:
Apologies, apparently I cannot remember what I said and you are indeed correct that we discussed it and I agreed to the direction.

Screenshot_20250517-075832

In hindsight I don't agree with it as stated above with fresher eyes, but I can't dispute it now.

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
```diff
@@ -126,7 +126,7 @@ internal static unsafe BOOL WritePrivateProfileString(string lpAppName, string?
             fixed (char* lpKeyNameLocal = lpKeyName)
             fixed (char* lpAppNameLocal = lpAppName)
             {
-                var res = PInvoke.WritePrivateProfileString(lpAppNameLocal, !string.IsNullOrWhiteSpace(lpKeyName) ? lpKeyNameLocal : null, !string.IsNullOrWhiteSpace(lpString) ? lpStringLocal : null, lpFileNameLocal);
+                var res = PInvoke.WritePrivateProfileString(lpAppNameLocal, !string.IsNullOrWhiteSpace(lpKeyName) ? lpKeyNameLocal : null, (lpString == null) ? null : lpStringLocal, lpFileNameLocal);
```
mjr4077au, line 129: This should just use the null coalescing operator.

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
```diff
@@ -76,9 +81,18 @@ function Remove-ADTIniSectionKey
...
+                if ($null -eq $Key)
+                {
+                    Write-ADTLogEntry -Message "Removing INI section: [$Section]."
+                }
+                else
+                {
+                    Write-ADTLogEntry -Message "Removing INI value: [Section = $Section] [Key = $Key]."
+                }
+
+                # If $Key is null, it will remove the entire section, otherwise it will remove just the key. [NullString]::Value is used as $null was being interpreted as an empty string.
```
mjr4077au, line 93: So based on your findings, what's stopping `$Key` from being interpreted as an empty string here?

mjr4077au (changes requested):
Once these comments are addressed, it can be merged in.

Issue #1464, opened 2025-05-19
[Bug] Translatable string review


alexhass:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

dev

### Describe the bug

I'm currently looking at the translatable strings to improve the German translation. As translator I found strings that look difficult to translate. I have not fully reviewed the module - just the strings - so I cannot say where they are used and if there is a string concatenation used. String concatenation are very problematic for translations.

These sentences look like they are not the start of a sentence, is that correct? If they are the beginning of a sentence, they should be ucfirst. Otherwise we have context issue.

e.g. 'complete. A reboot is required.' may be "'Foo complete. A reboot is required.'". In this case the translatable string should use placeholder to allow context senstive translation in many languages. e.g. 'Installation of :product complete. A reboot is required.'. I also this should be "completed".

```
BalloonText = @{
        # Text displayed in the balloon tip for successful completion of a deployment type.
        Complete = 'complete.'

        # Text displayed in the balloon tip for a failed deployment type.
        Error = 'failed.'

        # Text displayed in the balloon tip for fast retry of a deployment.
        FastRetry = 'not complete.'

        # Text displayed in the balloon tip for successful completion of a deployment type.
        RestartRequired = 'complete. A reboot is required.'

        # Text displayed in the balloon tip for the start of a deployment type.
        Start = 'started.'
    }
```

I'm not sure where it is used, but strings like these could cause translation issues if used in text.

- `TimeRemaining = 'Time remaining:'` -> `TimeRemaining = 'Time remaining: {time}` where {time} is a placeholder for the time string.
- 'Remaining Deferrals:' -> 'Remaining deferrals: {deferrals}'

I'm not a native english speaker, but in translation projects I participate we typically write:

- 'Restart Now' -> 'Restart now'
- 'Restart Required' -> 'Restart required'
- 'Close Apps & Install' -> 'Close apps & install'
- 'Close &Programs' -> 'Close &programs'

If placeholders are used it is very good to use named placeholder as it is not clear for translators what {0} may mean in the context. Is ist a name? Than may use {name}

[...]

Issue #1373, opened 2025-03-10
[Bug]  UI functions only using the -Defer parameter should not show "(1 time)" if there are no other defer parameters


Robosx:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.0.6

### Describe the bug

When UI functions use the “-AllowDefer” parameter and no other defer parameters, the defer button always shows “Defer (1 remain)” but it should just show “Defer”.

### Steps to reproduce

1. Use "Show-ADTInstallationWelcome -Defer" with no other defer parameters. 
2.
3.

### Environment data

```console
Windows 11
```

UraniumDaredevil:
Similar issue for me. I am using "-DeferDays" to set a future deadline date but every time the window shows up it displays 1 deferent remaining.

tjgruber:
@mjr4077au this bug introduced from https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1051

Also happens if only `-DeferTimes 7` is specified. I think the `ADTWelcomePromptClassic/Fluent` could benefit from additional parameters to account for more scenarios such as these.

mjr4077au:
This has been addressed within a work branch.

PR #1465, opened 2025-05-19
Update German translation

alexhass:
Fixing strings for case #1464

### DIFF
--- a/src/PSAppDeployToolkit/Strings/de/strings.psd1
+++ b/src/PSAppDeployToolkit/Strings/de/strings.psd1
@@ -1,29 +1,29 @@
 ﻿@{
     BalloonText = @{
         Complete = @{
-            Install = 'Installation abgeschlossen.'
-            Repair = 'Reparatur abgeschlossen.'
-            Uninstall = 'Deinstallation abgeschlossen.'
+            Install = 'Installation wurde abgeschlossen.'
+            Repair = 'Reparatur wurde abgeschlossen.'
+            Uninstall = 'Deinstallation wurde abgeschlossen.'
         }
         Error = @{
-            Install = 'Installation fehlgeschlagen.'
-            Repair = 'Reparatur fehlgeschlagen.'
-            Uninstall = 'Deinstallation fehlgeschlagen.'
+            Install = 'Installation ist fehlgeschlagen.'
+            Repair = 'Reparatur ist fehlgeschlagen.'
+            Uninstall = 'Deinstallation ist fehlgeschlagen.'
         }
         FastRetry = @{
-            Install = 'Installation nicht abgeschlossen.'
-            Repair = 'Reparatur nicht abgeschlossen.'
-            Uninstall = 'Deinstallation nicht abgeschlossen.'
+            Install = 'Installation wurde nicht abgeschlossen.'
+            Repair = 'Reparatur wurde nicht abgeschlossen.'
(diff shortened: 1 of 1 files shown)


src/PSAppDeployToolkit/Strings/de/strings.psd1
```diff
@@ -125,22 +125,22 @@
...
+                Uninstall = 'PSAppDeployToolkit - Deinstallieren der Anwendung'
             }
             DialogMessage = 'Bitte speichern Sie Ihre Arbeit, bevor Sie fortfahren, da die folgenden Anwendungen automatisch geschlossen werden.'
             DialogMessageNoProcesses = @{
                 Install = 'Wählen Sie Installieren aus, um mit der Installation fortzufahren. Wenn Sie noch Aufschübe haben, können Sie die Installation auch verschieben.'
                 Repair = 'Wählen Sie Reparieren aus, um mit der Reparatur fortzufahren. Wenn Sie noch Aufschübe haben, können Sie auch die Reparatur aufschieben.'
                 Uninstall = 'Wählen Sie Deinstallieren aus, um mit der Deinstallation fortzufahren. Wenn Sie noch Aufschübe haben, können Sie auch die Deinstallation aufschieben.'
             }
-            ButtonDeferRemaining = 'bleiben'
+            ButtonDeferRemaining = 'Beibehalten'
```
neo42JBR, line 138: This one is erronous and should be `verbleibend`

alexhass: What should that intentionally mean? There is no context and this makes it impossible to understand.

neo42JBR: This text is a button text that starts with the amount of times button can be pressed.
Sth like. `5 bleiben` or `5 verbleibend`
The suggested change `5 Beibehalten` would not make sense in this case.

I think with #1373 will change this string in the near future

alexhass: I agree, this make sense. Thanks for the details.

This means the english string has context sensitive translation bugs than. It should be “{number} verbleiben”. 

The english source string should be improved.

mjr4077au:
@alexhass I'm going to hold off on this for the time being. The strings.psd1 file I linked to you last night wasn't from this branch, but `dialogs-rework3`. There's still some further work to do before I bring that branch into develop. Once that's done though, I'd welcome an update to this PR.

Regrettably, I'm my no means a polyglot so all I have at my disposal is AI translations. However, once 4.1 is out, there'll be a lot more string file stability so improved translations from all languages would be welcome.

alexhass:
The patch is done against "develop" branch that has the latest commits. It was not against dialogs-rework3. Maybe we can get the fixes into the final 4.1, but I need to know when it's the best moment to update.

mjr4077au:
You can either redo this against `dialogs-rework3` now, or `develop` once `dialogs-rework3` is merged into it.

Using Windows release names, consider `develop` as our insider preview and `dialogs-rework3` as our canary branch

mjr4077au:
Merged manually via https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/a23ed5b0e115e405e01b6d9c9822227f180c6a1a. Done as a cherry-pick to ensure attribution is correct.

PR #1483, opened 2025-06-09
*-ADTIniSection commands

DanGough:
- Add new commands to Get/Set/Remove entire INI sections (Set accepts IDictionary, Get returns an ordered hashtable).
- Remove-ADTIniValue can no longer remove entire sections (i.e. back to how it was in 4.0)

We can incorporate these changes into develop now, additional enhancements can come in future, e.g:

- Merging of INI sections/contents
- Ability to process for all users
- Provide similar capabilities for JSON/YAML/XML


src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
```diff
@@ -85,6 +85,51 @@ internal static FARPROC GetProcAddress(SafeHandle hModule, string lpProcName)
...
+        /// <param name="lpFileName"></param>
+        /// <returns></returns>
+        /// <exception cref="Win32Exception"></exception>
+        /// <exception cref="OverflowException"></exception>
+        internal static uint GetPrivateProfileSectionNames(Span<char> lpReturnedString, string lpFileName)
+        {
+            var res = PInvoke.GetPrivateProfileSectionNames(lpReturnedString, lpFileName);
+            if (res == 0)
+            {
+                throw ExceptionUtilities.GetExceptionForLastWin32Error();
```
mjr4077au: The P/Invoke of `GetPrivateProfileSectionNames()` does not have `SetLastError=true`, so this check should be removed and just the `if (res == lpReturnedString.Length - 1)` test remaining.

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
```diff
@@ -85,6 +85,51 @@ internal static FARPROC GetProcAddress(SafeHandle hModule, string lpProcName)
...
+        /// <param name="lpFileName"></param>
+        /// <returns></returns>
+        /// <exception cref="Win32Exception"></exception>
+        /// <exception cref="OverflowException"></exception>
+        internal static uint GetPrivateProfileSection(string lpAppName, Span<char> lpReturnedString, string lpFileName)
+        {
+            var res = PInvoke.GetPrivateProfileSection(lpAppName, lpReturnedString, lpFileName);
+            if (res == 0)
+            {
+                throw ExceptionUtilities.GetExceptionForLastWin32Error();
```
mjr4077au: The P/Invoke of `GetPrivateProfileSection()` does not have `SetLastError=true`, so this check should be removed and just the `if (res == lpReturnedString.Length - 1)` test remaining.

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
```diff
@@ -110,6 +155,23 @@ internal static uint GetPrivateProfileString(string lpAppName, string lpKeyName,
...
+        /// <param name="lpAppName"></param>
+        /// <param name="lpString"></param>
+        /// <param name="lpFileName"></param>
+        /// <exception cref="Win32Exception"></exception>
+        internal static BOOL WritePrivateProfileSection(string lpAppName, string? lpString, string lpFileName)
+        {
+            var res = PInvoke.WritePrivateProfileSection(lpAppName, lpString, lpFileName);
+            if (!res)
+            {
+            throw ExceptionUtilities.GetExceptionForLastWin32Error();
```
mjr4077au: Missing indentation here.

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
```diff
@@ -127,11 +189,11 @@ internal static unsafe BOOL WritePrivateProfileString(string lpAppName, string?
             fixed (char* lpAppNameLocal = lpAppName)
             {
                 var res = PInvoke.WritePrivateProfileString(lpAppNameLocal, !string.IsNullOrWhiteSpace(lpKeyName) ? lpKeyNameLocal : null, (lpString != null) ? lpStringLocal : null, lpFileNameLocal);
-                if (!res)
-                {
-                    throw ExceptionUtilities.GetExceptionForLastWin32Error();
-                }
-                return res;
```
mjr4077au, line 134: Revert these indent changes here.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -1,6 +1,10 @@
-﻿using System;
-using PSADT.Extensions;
+﻿using PSADT.Extensions;
 using PSADT.LibraryInterfaces;
+using System;
+using System.Collections;
```
mjr4077au, line 2: Preferably, our using statements should be below the System namespace ones.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+        /// <returns>OrderedDictionary of key/value pairs in the section</returns>
+        public static OrderedDictionary? GetSection(string section, string filepath)
+        {
+            var sections = GetSectionNames(filepath);
+            if (!sections.Contains(section, StringComparer.OrdinalIgnoreCase))
+            {
+                throw new ArgumentException($"Section [{section}] was not found in the INI file.");
+            }
+
+            var buffer = new char[65536];
```
mjr4077au: Do `Span<char> = stackalloc char[65536]` here so the memory is allocated on the stack instead of the heap. 64KB is a large-ish allocation, but I don't believe the callstack would be deep enough to cause a StackOverflowException.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            {
+                return null;
+            }
+
+            var entries = new string(buffer, 0, (int)res).Split('\0');
+            var dictionary = new OrderedDictionary();
+
+            foreach (var entry in entries)
+            {
+                if (string.IsNullOrEmpty(entry)) continue;
```
mjr4077au: Use `string.IsNullOrWhitespace()` as that's what we use exclusively throughout the rest of the assembly.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            var buffer = new char[65536];
+
+            uint res;
+            try
+            {
+                res = Kernel32.GetPrivateProfileSection(section, buffer, filepath);
+            }
+            catch
+            {
+                return null;
```
mjr4077au: Don't swallow errors here as it's a silent failure. This should throw an exception for handling on the PowerShell side, with the caller deciding whether it's an error or not by using `-ErrorAction` against their function call.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            }
+            if (res == 0)
+            {
+                return null;
```
mjr4077au: Same here. If we get nothing, we should throw an InvalidOperationException or something else that's suitable.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+        /// </summary>
+        /// <param name="section">The section name</param>
+        /// <param name="filepath">Path to the INI file</param>
+        /// <returns>OrderedDictionary of key/value pairs in the section</returns>
+        public static OrderedDictionary? GetSection(string section, string filepath)
+        {
+            var sections = GetSectionNames(filepath);
+            if (!sections.Contains(section, StringComparer.OrdinalIgnoreCase))
+            {
+                throw new ArgumentException($"Section [{section}] was not found in the INI file.");
```
mjr4077au: This should be `throw new ArgumentException($"Section [{section}] was not found in the INI file.", nameof(section));`.

src/PSADT/PSADT/NativeMethods.txt
```diff
@@ -37,6 +37,8 @@ GetExitCodeProcess
 GetExitCodeThread
 GetForegroundWindow
 GetNativeSystemInfo
+GetPrivateProfileSectionNames
```
mjr4077au: This should be under `GetPrivateProfileSection`, I believe.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            catch
+            {
+                return null;
+            }
+            if (res == 0)
+            {
+                return null;
+            }
+
+            var entries = new string(buffer, 0, (int)res).Split('\0');
```
mjr4077au: When you use `stackalloc` for the buffer, this can be used straight into a `Span<char>` variable. Once that's done, use `buffer.Slice()` here.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            var dictionary = new OrderedDictionary();
+
+            foreach (var entry in entries)
+            {
+                if (string.IsNullOrEmpty(entry)) continue;
+
+                var separatorIndex = entry.IndexOf('=');
+                if (separatorIndex <= 0) continue;
```
mjr4077au, line 77: Should we consider throwing here? Is it indicative of a malformed INI file?

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+
+                var key = entry.Substring(0, separatorIndex);
+                var value = entry.Substring(separatorIndex + 1);
+
+                if (dictionary.Contains(key))
+                {
+                    dictionary[key] = value;
```
mjr4077au, line 84: Should we really just be overwriting existing values here? What happens in situations where there's multiple key/values with the same key, but different casing? If `OrderedDictionary` can't handle that, then `Dictionary<string, string>` should be used instead, though I do understand why you went with `OrderedDictionary` since it's more familiar and usable for PowerShell folk who aren't on PowerShell 7.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+        }
+
+        /// <summary>
+        /// Gets all section names from an INI file.
+        /// </summary>
+        /// <param name="filepath">Path to the INI file</param>
+        /// <returns>Array of section names</returns>
+        public static string[] GetSectionNames(string filepath)
+        {
+            var buffer = new char[65536];
```
mjr4077au: As per the above, use `Span<char> = stackalloc char[65536]` here.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            var res = Kernel32.GetPrivateProfileSectionNames(buffer, filepath);
+            var names = new string(buffer, 0, (int)res).Split('\0').Where(name => !string.IsNullOrEmpty(name)).ToArray();
```
mjr4077au: Do `string.IsNullOrWhiteSpace()` here.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+
+            return dictionary;
+        }
+
+        /// <summary>
+        /// Gets all section names from an INI file.
+        /// </summary>
+        /// <param name="filepath">Path to the INI file</param>
+        /// <returns>Array of section names</returns>
+        public static string[] GetSectionNames(string filepath)
```
mjr4077au: Since this is only used internally, make this private or just remove the method entirely and bake it into the above method.

mjr4077au: For all functions, it's preferable that they return their data as readonly where possible and it makes sense to do so. If this remains public, it should be returning `IReadOnlyList<string>`. If it's privatised, make it return `ReadOnlyCollection<string>`.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
mjr4077au: If the signature of this method is changed to return readonly data, this should change from `.ToArray()` to `.ToList().AsReadOnly()`.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+        public static void WriteSection(string section, IDictionary? keyValuePairs, string filepath)
+        {
+            if (keyValuePairs == null)
+            {
+                Kernel32.WritePrivateProfileSection(section, null, filepath);
+                return;
+            }
+
+            var entries = new StringBuilder();
+            if (keyValuePairs.Count == 0)
```
mjr4077au: Try to put the most likely scenario in the if statement and the exception case in the else statement. I feel it's more likely that a populated dictionary will be passed.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            return names;
+        }
+
+        /// <summary>
+        /// Writes multiple key/value pairs to a section in an INI file.
+        /// </summary>
+        /// <param name="section">The section name</param>
+        /// <param name="keyValuePairs">Collection of key/value pairs to write</param>
+        /// <param name="filepath">Path to the INI file</param>
+        public static void WriteSection(string section, IDictionary? keyValuePairs, string filepath)
```
mjr4077au: The name `keyValuePairs` implies the parameter takes a different type. `Hashtable` and `OrderedDictionary` objects are comprised of `DictionaryEntry` objects. A true `Dictionary<TKey, TValue>` is comprised of `KeyValuePair<TKey, TValue>` objects. Maybe simply name it `data`?

mjr4077au: Would it be more logical that this method be ordered from `public static void WriteSection(string filepath, string section, IDictionary? data)`? I appreciate it's ordered the way the underlying P/Invoke is, but we don't necessarily need to honour that.

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
```diff
@@ -0,0 +1,105 @@
...
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
+            })]
+        [System.String]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Section`.

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
```diff
@@ -64,11 +64,23 @@ function Get-ADTIniValue
         [System.String]$FilePath,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Section`.

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
```diff
@@ -0,0 +1,105 @@
...
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
```
mjr4077au: Just return `$true` here.

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
```diff
@@ -64,11 +64,23 @@ function Get-ADTIniValue
...
+                return $true
+            })]
         [System.String]$Section,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified key cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Key`.

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName FilePath -ProvidedValue $_ -ExceptionMessage 'The specified file does not exist.'))
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
+            })]
+        [System.String]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Remove-ADTIniSection.ps1
```diff
@@ -0,0 +1,102 @@
...
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
+            })]
+        [System.String]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Section`.

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
```diff
@@ -64,11 +64,23 @@ function Get-ADTIniValue
         [System.String]$FilePath,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
```diff
@@ -64,11 +64,23 @@ function Get-ADTIniValue
...
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return $true
+            })]
         [System.String]$Section,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Remove-ADTIniSection.ps1
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName FilePath -ProvidedValue $_ -ExceptionMessage 'The specified file does not exist.'))
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
+            })]
+        [System.String]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
```diff
@@ -62,11 +62,23 @@ function Remove-ADTIniValue
         [System.String]$FilePath,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
```diff
@@ -62,11 +62,23 @@ function Remove-ADTIniValue
...
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return $true
+            })]
         [System.String]$Section,
 
         [Parameter(Mandatory = $false)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
+            })]
+        [string]$Section,
+
+        [Parameter(Mandatory = $true)]
+        [AllowNull()]
```
mjr4077au, line 90: Should this also have `[AllowEmptyCollection()]` as well?

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
+            })]
+        [string]$Section,
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+        [ValidateNotNullOrEmpty()]
+        [string]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
```
mjr4077au: Just return `$true` here.

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
```diff
@@ -0,0 +1,105 @@
...
+        Copyright: (C) 2025 PSAppDeployToolkit Team (Sean Lillis, Dan Cunningham, Muhammad Mashwani, Mitch Richters, Dan Gough).<br />
+        License: https://opensource.org/license/lgpl-3-0
+
+    .LINK
+        https://psappdeploytoolkit.com/docs/reference/functions/Get-ADTIniValue
+    #>
+
+    [CmdletBinding()]
+    [OutputType([Collections.Specialized.OrderedDictionary])]
+    param (
```
mjr4077au: This opening bracket should be on a new line.

src/PSAppDeployToolkit/Public/Remove-ADTIniSection.ps1
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return ![System.String]::IsNullOrWhiteSpace($_)
```
mjr4077au: Just return `$true` here.

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
```diff
@@ -62,11 +62,23 @@ function Remove-ADTIniValue
         [System.String]$FilePath,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Section`.

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
```diff
@@ -62,11 +62,23 @@ function Remove-ADTIniValue
...
+                return $true
+            })]
         [System.String]$Section,
 
         [Parameter(Mandatory = $false)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified key cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Key`.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+        Website: https://psappdeploytoolkit.com<br />
+        Copyright: (C) 2025 PSAppDeployToolkit Team (Sean Lillis, Dan Cunningham, Muhammad Mashwani, Mitch Richters, Dan Gough).<br />
+        License: https://opensource.org/license/lgpl-3-0
+
+    .LINK
+        https://psappdeploytoolkit.com/docs/reference/functions/Set-ADTIniSection
+    #>
+
+    [CmdletBinding()]
+    param (
```
mjr4077au: This opening bracket should be on a new line.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$FilePath,
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+        License: https://opensource.org/license/lgpl-3-0
+
+    .LINK
+        https://psappdeploytoolkit.com/docs/reference/functions/Set-ADTIniSection
+    #>
+
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
```
mjr4077au, line 76: You've used a `[ValidateScript()]` decoration for every other `-FilePath` parameter, but not this one.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+        [string]$FilePath,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Section`.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+                            TargetObject = $FilePath
+                            RecommendedAction = "Please confirm the path of the specified file and try again, or add -Force to create a new file."
+                        }
+                        throw (New-ADTErrorRecord @naerParams)
+                    }
+                    Write-ADTLogEntry -Message "Creating INI file: $FilePath."
+                    $null = New-Item -Path $FilePath -ItemType File -Force
+                }
+
+                if ($null -eq $Content -or $Content.Count -eq 0)
```
mjr4077au: Maybe require `-Force` with an empty dictionary as well, just so people don't accidentally obliterate their INI file.

src/PSAppDeployToolkit/Public/Set-ADTIniValue.ps1
```diff
@@ -63,11 +63,23 @@ function Set-ADTIniValue
         [System.String]$FilePath,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Section`.

src/PSAppDeployToolkit/Public/Set-ADTIniValue.ps1
```diff
@@ -63,11 +63,23 @@ function Set-ADTIniValue
...
+                return $true
+            })]
         [System.String]$Section,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
+                {
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified key cannot be null, empty, or whitespace.'))
```
mjr4077au: `-ParameterName Path` should be `-ParameterName Key`.

src/PSAppDeployToolkit/Public/Set-ADTIniValue.ps1
```diff
@@ -63,11 +63,23 @@ function Set-ADTIniValue
...
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified section cannot be null, empty, or whitespace.'))
+                }
+                return $true
+            })]
         [System.String]$Section,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Set-ADTIniValue.ps1
```diff
@@ -63,11 +63,23 @@ function Set-ADTIniValue
         [System.String]$FilePath,
 
         [Parameter(Mandatory = $true)]
-        [ValidateNotNullOrEmpty()]
+        [ValidateScript({
+                if ([string]::IsNullOrWhiteSpace($_))
```
mjr4077au: You're using a mix of `[string]` and `[System.String]`. While I prefer the latter and it's the more common vernacular within the module, whatever you choose at least be consistent.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
```diff
@@ -0,0 +1,151 @@
...
+                [PSADT.Utilities.IniUtilities]::WriteSection($Section, $Content, $FilePath)
+            }
+            catch
+            {
+                Write-Error -ErrorRecord $_
+            }
+        }
+        catch
+        {
+            Invoke-ADTFunctionErrorHandler -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState -ErrorRecord $_ -LogMessage "Failed to write INI value."
```
mjr4077au: Failed to write INI "value" or "section"?

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
mjr4077au: As mentioned above in another area of the review, I'd put the most likely scenario in the if statement, which in this case to me would be a populated dictionary.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            if (keyValuePairs.Count == 0)
+            {
+                entries.Append('\0');
+            }
+            else
+            {
+                foreach (DictionaryEntry kvp in keyValuePairs)
+                {
+                    string key = kvp.Key?.ToString()?.Trim() ?? string.Empty;
+                    if (string.IsNullOrEmpty(key))
```
mjr4077au: Use `string.IsNullOrWhiteSpace()` here.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+            var entries = new StringBuilder();
+            if (keyValuePairs.Count == 0)
+            {
+                entries.Append('\0');
+            }
+            else
+            {
+                foreach (DictionaryEntry kvp in keyValuePairs)
+                {
+                    string key = kvp.Key?.ToString()?.Trim() ?? string.Empty;
```
mjr4077au: Do INI files accept empty key values? I feel a null key would be an exception-causing event because it'd be a bad setup from the caller.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
mjr4077au: As said above about why I understand the bulk of the code is using `OrderedDictionary`, I understand why this is `IDictionary` also for the same reason. Just be mindful that an `IDictionary` like a `Hashtable` or `OrderedDictionary` is effectively `Dictionary<object, object>`. As such, there's nothing stopping the caller from sending through complex values like hashtables, arrays, or other types that may not necessarily be representable as a string. It would be more technically correct to use `Dictionary<string, string>` for all of this code.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
```diff
@@ -34,5 +38,115 @@ public static void WriteSectionKeyValue(string section, string? key, string? val
...
+        /// </summary>
+        /// <param name="filepath">Path to the INI file</param>
+        /// <returns>Array of section names</returns>
+        public static string[] GetSectionNames(string filepath)
+        {
+            var buffer = new char[65536];
+            var res = Kernel32.GetPrivateProfileSectionNames(buffer, filepath);
+            var names = new string(buffer, 0, (int)res).Split('\0').Where(name => !string.IsNullOrEmpty(name)).ToArray();
+
+            return names;
```
mjr4077au: There's no need to store `names` as a variable, only to just return it on the next line. Just return it directly.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Intentional, same as Set-ADTIniValue, since Force can be used to create the file if it does not exist.

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniSection.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Remove-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
DanGough: Resolved

src/PSAppDeployToolkit/Public/Get-ADTIniValue.ps1
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
DanGough: Resolved

src/PSADT/PSADT/NativeMethods.txt
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Will keep it private for now as future ini functions might rely on it.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Agreed, I'll make it throw here if anyone gives it crazy input like that. Trying to set a null/empty/whitespace value via WritePrivateProfileString results in the value being deleted, so we should consider it invalid.

src/PSADT/PSADT/LibraryInterfaces/PSADT.LibraryInterfaces.Class.Kernel32.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: It's valid to have an empty section, so I wouldn't want to throw here. Right now it returns null for an empty section, but maybe an empty dictionary would be better, which we can get by just removing this check entirely.

Then you could retrieve a section, add content to it, then write it somewhere without having to worry if the original content was empty.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: I've done some further testing and the GetPrivateProfileSection ignores any lines beginning with ; to denote a comment. However, other INI implementations can use # to denote a comment, and GetPrivateProfileSection will bring these back.

So while technically they may indicate a malformed INI file, I wouldn't want to bomb out just in case we come across one of these unexpected lines, so I'll leave this bit as-is.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: If you try and use WritePrivateProfileString to an entry in lowercase, then again to the same entry name in uppercase, it ignore the case and overwrites the original entry.

So from that we can assume that multiple keys with different casing is not valid/supported. Then it's reasonable to assume that last entry wins, which is what this setup would produce.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Renamed to content, to match the name used on the PowerShell side of things.

DanGough: True, happy to switch this up as long as the users can send data as a hashtable @{'Like' = 'This'}, without forcing them to create a [System.Collections.Generic.Dictionary[string,string]] object they may have never have heard of.

I'll see if I can enforce Dictionary<string, string> but retain the flexibility on the PowerShell side so that it can massage the input into that form before calling the C# function.

DanGough: We need to maintain the input order, so will stick to the OrderedDictionary, but sanity check the inputs instead.

DanGough: Agreed, that would mean changing the other ini functions to suit, but that's no big deal.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Resolved

src/PSAppDeployToolkit/Public/Set-ADTIniSection.ps1
DanGough: I think we can leave it as-is, as it introduces double purposes for -Force. They might also accidentally erase data by sending a section with a single value if they don't know what they're doing. I'll add extra comments to the help.

src/PSADT/PSADT/Utilities/PSADT.Utilities.Class.IniUtilities.cs
DanGough: Done.

PR #1487, opened 2025-06-12
Change default msiexec.exe parameters from `/qb-!` to `/qn`

mjr4077au:
The previous value, while cool by showing a msiexec.exe progress bar without the ability to cancel/stop in any way, is not always properly honoured by vendors who can still display modal dialogs in this mode. This is only truly addressed by using `/qn` to fully silence a deployment.


DanGough (approved):
Looks good to me, I for one always made this change anyway in my own templates.

mjr4077au:
@DanGough you need to also amend the ADMX template descriptions also, doing just the config.psd1 file is not enough.

PR #1488, opened 2025-06-13
Change switch `-AllowRebootPassThru` to `-SuppressRebootPassThru`

mjr4077au:
The way this was isn't the most ideal as it's natural to want to allow a reboot code to go back through to your deployment mechanism. Ideally this switch shouldn't exist at all, but we supply the antonym for those who were relying on it. The right solution is to just control it in SCCM or Intune on a per-app basis.


DanGough (commented):
Agreed this should normally be controlled at the deployment side by setting 3010 to mean 'Success', but having the switch there is still a nice option to have.

The only time I had to rely on PSADT masking 3010 and returning 0 instead though was when deploying an app via a task sequence whilst the user was logged - in that scenario a 3010 resulted in an instant reboot. I may have been able to solve this differently by managing the return codes, but did not think of that at the time so simply remove the -AllowRebootPassThru switch.

My only concern is that the toolkit now fails to start if we supply -AllowRebootPassThru, since this passes the arg direct to the ps1 which does not recognise it. This is obvious when running the new /debug switch on Invoke-AppDeployToolkit.exe:

image

However this info never made it to the log:

image

Looks like Invoke-AppDeployToolkit.exe logs should be enhanced to also capture StdError here?

If we stick to this behaviour rather than ignoring that parameter, we'll have to make sure it is documented and broadcast loud and clear.

mjr4077au:
We can't capture stdout/stderr in exe's log file without removing the automatic UAC elevation if RequireAdmin is true. Basically we'd have to choose one or the other.

JonathanPitre:
Glad to see a /debug option added!

mjr4077au:
@JonathanPitre the upcoming 4.1.0 release is going to be a true red-letter day for the project 🤘

PR #1507, opened 2025-07-02
Update strings.psd1 (sv)

ClientMGMT:
Fixed several translation errors for dialog message text in Swedish (sv).


mjr4077au:
Thank you boss 🤘

PR #1517, opened 2025-07-07
revise German UI strings for neutral tone

lukashaensel:
Removed all explicit “Sie/du” pronouns to stay address-agnostic


mjr4077au:
Please rework this against our develop branch and not main.

PR #1569, opened 2025-07-31
Update strings.psd1

ClientMGMT:
Fluent / DialogMessageNoProcesses
Remove the following text from both the Repair and Uninstall messages (the last sentence): 'If you have any deferrals remaining, you may also choose to delay the repair.'

This change will make the default messages be in sync with how the default template is configured, no deferrals are configured for Repair & Uninstall in Invoke-AppDeployToolkit.ps1. Defer is only configured for the Install function by default and thereby the additional sentence 'If you have any deferrals remaining, you may also choose to delay the repair.' should only exist in the Install message of 'DialogMessageNoProcesses'.

If you need to verify this in your own tests do the following: Add a process that needs to be closed to your package and make sure it's started. Run an install/repair/uninstall of the package, when it stops and asks you to Save and Close, instead of clicking the save and close button just close the process/app instead, this will change the message text to the 'DialogMessageNoProcesses' one.

PS. This is already fixed in the SV strings.psd1 file but all other language 'strings' needs to be updated as well.

### DIFF
--- a/src/PSAppDeployToolkit/Strings/strings.psd1
+++ b/src/PSAppDeployToolkit/Strings/strings.psd1
@@ -200,8 +200,8 @@
             # This is a message to when there are no running processes available.
             DialogMessageNoProcesses = @{
                 Install = 'Please select Install to continue with the installation. If you have any deferrals remaining, you may also choose to delay the installation.'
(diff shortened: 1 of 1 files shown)


ClientMGMT:
That change is even better 👍

sintaxasn (approved):

PR #1598, opened 2025-08-12
Update strings.psd1 (for Finnish language)

labradmin:
DialogMessageNoProcesses, the text has Install and Repair in English instead of Finnish. Changed "Install" to "Asenna" and "Repair" to "Korjaa" (Since the buttons say this anyway).


sintaxasn (approved):
Thanks for the commit! :)

PR #1711, opened 2025-09-04
Fix `OutOfMemoryException` issues within `Resolve-ADTErrorRecord`.

mjr4077au:
There's bugs in `Out-String` that can lead to strings with infinitely long trailing lines depending on how wide the output is to be. We resolve this by streaming the output to reduce memory load, and by reducing the width maximum 16-bit value, as we don't need 2147483647 columns here.


DanGough (approved):

DanGough:
This can go in 4.1.x also, if we issue a 4.1.5 before 4.2.0...

PR #1738, opened 2025-09-10
Ensure all callbacks removed via `Remove-ADTModuleCallback` are referenced by the CommandTable entry.

mjr4077au:
The result of `$MyInvocation.MyCommand` is referentially different to what's obtained via `Get-Command` when used to populate the CommandTable during module import.


sintaxasn (approved):
Makes sense

Issue #1639, opened 2025-08-19
[Feature] Multiple text formatings & url shortening


AlkHacNar:
### Summary of the new feature / enhancement

With the 4.1.0 update its now possible to use bold, italic, assent und urkl formatting in the Show-ADTInstallationWelcome in the custom text field, and probably on more commands, but I just used this one.. But as of now, or I didn't figure out how, you can only use one of them on a text, nore multiple. Would it be possible to implement, that I could do m[bold][italic] or something like this? And the second thing ist, would it be possible to have the [url] to shorten the link, like [url="Link"]https://www.test.de[/url]. sharepoint urls are sometime very long ^^

sintaxasn:
Thanks for the suggestions. Funnily enough, I hadn't thought about combining multiple bbtags like bold AND italic. Doh! 🤦🏻

URL Text + Link is something we wanted to do but just didn't get finished in time. But is something we'll add in the next release (4.1.x or possibly 4.2.x)

cyborgrd:
Where i can see a manual, how format text inside of Show-ADTInstallationWelcome? BBCode ? which are supported?

AlkHacNar:
I think there isn't a manual, but you just use [bold]text to format[/bold]. Change bold with the others for other formats

PR #1749, opened 2025-09-15
Added support for Link Descriptions in URL tags, as well as nested tags

sintaxasn:
- Added support for providing a Link Description when specifying a URL tag. This is done using the same method as BBCode - the format is: [url=https://mysite.com]MySite Description[/url]. Addresses #1639 
- This allows for richer formatting options in dialogs, and remains backwards compatible with the existing [url]https://mysite.com[/url] tag.
- Updated text formatting logic to support nested and combined tags for accent, bold, and italic styles. Addresses second part of #1639 
- Brought back CustomText as a way to test each tag type and nested tags in the Test Harness.
- Fixed issue in Test Harness caused by introduction of Language parameter to Dialogs (to support RTL text).

### DIFF
--- a/src/PSADT/PSADT.UserInterface.TestHarness/Program.cs
+++ b/src/PSADT/PSADT.UserInterface.TestHarness/Program.cs
@@ -11,6 +11,7 @@ using PSADT.UserInterface.DialogOptions;
 using PSADT.UserInterface.DialogResults;
 using PSADT.UserInterface.Dialogs;
 using PSADT.UserInterface.DialogState;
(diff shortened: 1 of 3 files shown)


mjr4077au (changes requested):
Looks good for the Fluent dialog, however the code that strips these formatting tags out for the Classic dialogs will need some reworking prior to merging. Shouldn't take much to adjust.

image

sintaxasn:
Should now work in both scenarios.

mjr4077au (dismissed):

PR #1754, opened 2025-09-17
Test whether an app is an is an MSI app based on UninstallString, in addition to the WindowsInstaller bool.

mjr4077au:
Apparently there's MSI-based apps out there which don't set this correctly: https://discourse.psappdeploytoolkit.com/t/msi-uninstallation-not-working-in-v4-1-5/6947/7

### DIFF
--- a/src/PSAppDeployToolkit/Public/Get-ADTApplication.ps1
+++ b/src/PSAppDeployToolkit/Public/Get-ADTApplication.ps1
@@ -232,21 +232,13 @@ function Get-ADTApplication
                     }
 
                     # Build hashtable of calculated properties based on their presence in the registry and the value's validity.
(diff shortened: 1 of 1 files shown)


DanGough:
Confirmed that the specific package has a custom action that overwrites the Uninstall entry after the PublishProduct/InstallFinalize actions, so this is a rare edge case and not something potentially widespread.

mjr4077au:
@DanGough with that in mind, do we still wish to merge it? I don't see the harm of having this in as a fallback

DanGough (approved):

DanGough:
@mjr4077au should be fine, but it would be wise to test what happens if there's a name-matching Uninstall entry with msiexec present in the uninstallstring (e.g. msiexec /x C:\Temp\package.msi), but no valid ProductCode GUID in the registry key name or uninstallstring.

Issue #1756, opened 2025-09-17
[Bug] EXE wrapper strips commas from arguments before forwarding to Invoke-AppDeployToolkit.ps1


PierreHachard:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.5

### Describe the bug

Hey guys,

Thank you for all the work on the project and the recently new features added since the 4.1 release.

## Summary

When invoking Invoke-AppDeployToolkit.exe, a comma-separated value passed as an extra argument loses its commas before it reaches Invoke-AppDeployToolkit.ps1.
Calling the .ps1 directly preserves the commas as expected.

## Details

The steps to reproduce provide more details on the behaviour.

Not sure if this is intended, it could be some parsing volontarily implemented by the EXE wrapper, so I added a question label.

I have encountered this issue on both PSADT 4.1.3 and 4.1.5 (upgraded to 4.1.5 to specifically test this behavior).

Note that I could not reproduce the issue with PSADT 4.0.6, which is what we were using before the release of 4.1.X.

## Changes made to the Invoke-AppDeployToolkit.ps1

The only changes made to the script to test the issue is adding a new parameter '$ExtraArgs' just after the '$DisableLogging' in the Invoke-AppDeployToolkit.ps1 script defined with the ValueFromRemainingArguments property, like so:

```
[Parameter(ValueFromRemainingArguments = $true)]
[string[]]$ExtraArgs
```

The use case for this is to be able to add arguments that will be passed down to the final installer (ex : Adobe Reader in my example) without having to inject everything directly in the Invoke-AppDeployToolkit.ps1 script.

### Steps to reproduce

1. Open an elevated PowerShell console and navigate to the toolkit folder.
2. Run the EXE wrapper with the following command:
`.\Invoke-AppDeployToolkit.exe -DeploymentType Install -DeployMode NonInteractive /sAll /rs /msi /qn LANG_LIST="fr_FR,en_US,es_ES" EULA_ACCEPT=YES`
3. Observe the initialization line in the PSADT logs shows:

[...]

PR #1759, opened 2025-09-17
Optimise argument parsing lambdas in `PSADT.Invoke` solution.

mjr4077au:
Observed while working on #1756. Using static lambda delegates allows the compiler to generate more optimised code since it knows that the lambda doesn't require access to the local state.


DanGough (approved):
Seems fine, from reading up on this, it would generate a compile time error if trying to use any external variables in these static lambdas.

PR #1853, opened 2025-10-23
Update installation prompt message in Slovak

lumirjasny:
Following the analogy of having "action" word in the sentence in English and (Slovak) in brackets, like on the 2 below lines for actions Repair(line134) and Uninstall(line135).  The install should be the same.


mjr4077au:
Thanks, looks good!

Issue #1860, opened 2025-10-24
[Feature] Test for interactivity before allowing `/Debug` to work


mjr4077au:
### Summary of the new feature / enhancement

People blindly adding this onto the command line could end up shooting themselves in the foot.

### Proposed technical implementation details (optional)

Test `Environment.UserInteractive` before allowing the switch to work.

Issue #1854, opened 2025-10-24
[Bug] Open-ADTSession fails on Systems with active culture tr-TR


SchmitzB:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.7

### Describe the bug

On Turkish Windows 11 systems that have to culture tr-TR active cannot correctly run PSADT.
The error is present with Version 4.1.5, 4.1.6 & 4.1.7 - Other Versions were not tested.
After testing we discovered that it fails on Open-ADTSession with the following error:

```
Open-ADTSession : Verilen anahtar sözlükte yoktu.
At Invoke-AppDeployToolkit.ps1:341 char:19
+     $adtSession = Open-ADTSession @adtSession @iadtParams -PassThru
+                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OpenError: (0:Int32) [Open-ADTSession], KeyNotFoundException
    + FullyQualifiedErrorId : System.Collections.Generic.KeyNotFoundException,Open-ADTSession
```

### Steps to reproduce

1. Change Regional Format to tr-TR
2. Start Invoke-AppDeployToolkit.exe with /Debug parameter

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 24H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64 bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 26100.1.amd64fre.ge_release.240331-1435
OsLanguage           : tr-TR
OsMuiLanguages       : {tr-TR, en-US}
KeyboardLayout       : tr-TR
TimeZone             : (UTC+03:00) İstanbul
HyperVisorPresent    : True
CsPartOfDomain       : False
CsPCSystemType       : Mobile
```

mjr4077au:
I can't replicate this by manually forcing the toolkit onto `tr-TR`:

[...]

Issue #1841, opened 2025-10-17
[Feature] User UI Culture Variable


cyborgrd:
### Summary of the new feature / enhancement

I want a session variable that contains the language of the interactively logged-in user, as shown in the log:

[Initialization] :: The following locale was used to import UI messages from the strings.psd1 files: **[en-US]**.

Our system default is de-DE, but some users have changed their culture to en-US. The dialog is displayed correctly, but I want to check the language for something like this:

```
if ($uiculture  -match 'de-DE') {
	Start-ADTMsiProcess -Action Install -Transforms "$ARPSourceFilesPath\KeePass-2.59.mst" -FilePath "$ARPSourceFilesPath\KeePass-2.59.msi" -AdditionalArgumentList 'LANGUAGE="DE"'
	}
	else {
	Start-ADTMsiProcess -Action Install -Transforms "$ARPSourceFilesPath\KeePass-2.59.mst" -FilePath "$ARPSourceFilesPath\KeePass-2.59.msi"	-AdditionalArgumentList 'LANGUAGE="EN"'
	}
```

When I use $UICulture, it always returns de-DE, even if the user has changed their culture.

mjr4077au:
I'll take a look at this, but I think it's probably best that new variables be introduced to support this. I personally wouldn't necessarily want to install something on a PC just because Susan has her culture set to x whereas the system and any other user who logs on has it set to y.

cyborgrd:
We have one user systems.  But this Variable is also useful to set correct user settings (language) to registry or ini files.

AlkHacNar:
same as my #1772

Issue #1838, opened 2025-10-15
[Bug] Show-ADTInstallationWelcome: CloseApps Prompt not showing for user v2


Fredaker:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.5

### Describe the bug

This appears to be a reoccurrence of issue [#1716](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1716).

When a package is executed in the SYSTEM context while a standard (non-administrator) user is logged in, any processes listed in `$adtSession.AppProcessesToClose` that are also running under the SYSTEM context are not displayed on the welcome prompt. This functionality works as expected when you are logged in as an administrator.

This behavior was first observed during package repair testing, but further testing confirms it is also present during installations and uninstallations. In our use case, where the defer option was disabled, this bug resulted in an endless loop of splash screens during a repair action because the prompt to close the application never appeared.

For testing purposes, notepad.exe was used as the process to be closed.
A log is located below.

The issue is also reproducible in version 4.1.6-PreRelease.

```
<![LOG[-------------------------------------------------------------------------------]LOG]!><time="19:41:12.619+120" date="10-15-2025" component="Open-ADTSession" context="NT AUTHORITY\SYSTEM" type="1" thread="5456" file="C:\Temp\test\PSAppDeployToolkit\PSAppDeployToolkit.psm1">
<![LOG[[Initialization] :: [Test_Software_1.0_x64_EN_01] install started.]LOG]!><time="19:41:12.642+120" date="10-15-2025" component="Open-ADTSession" context="NT AUTHORITY\SYSTEM" type="1" thread="5456" file="C:\Temp\test\PSAppDeployToolkit\PSAppDeployToolkit.psm1">
[...]

PR #1862, opened 2025-10-25
Some C#-related fixes.

mjr4077au:
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1860.
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1838.
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1772.
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1841.
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1854.
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1805.

@DanGough please pay special attention to the changes implemented for #1838 and let me know if you feel this sufficiently addresses the issues. Right now if there's processes to close that the user can't close themselves as they're non-admin, etc, the toolkit will now just close them.

Ultimately this is no different to 3.x and 4.0.x where the user to receive the dialog had the power to close other user's processes, but it still feels overarching.

### DIFF
--- a/src/PSAppDeployToolkit/Public/Show-ADTInstallationWelcome.ps1
+++ b/src/PSAppDeployToolkit/Public/Show-ADTInstallationWelcome.ps1
@@ -1073,6 +1073,7 @@ function Show-ADTInstallationWelcome
                     $dialogOptions = [PSADT.UserInterface.DialogOptions.CloseAppsDialogOptions]::new($DeploymentType, $dialogOptions)
 
                     # Spin until apps are closed, countdown elapses, or deferrals are exhausted.
+                    $sessionClosed = $false
                     while (($runningApps = Get-ADTRunningProcessesUserCanClose) -or (($promptResult -ne 'Defer') -and ($promptResult -ne 'Close')))
                     {
                         # Check if we need to prompt the user to defer, to defer and close apps, or not to prompt them at all
@@ -1095,7 +1096,7 @@ function Show-ADTInstallationWelcome
                                         if ($deferRunIntervalNextTime -gt [System.TimeSpan]::Zero)
                                         {
                                             Write-ADTLogEntry -Message "Next run interval not due until [$(($currentDateTimeLocal + $deferRunIntervalNextTime).ToString('O'))], exiting gracefully."
-                                            Close-ADTSession -ExitCode $adtConfig.UI.DefaultExitCode
+                                            $sessionClosed = $true; Close-ADTSession -ExitCode $adtConfig.UI.DefaultExitCode
                                         }
                                     }
                                 }
@@ -1200,12 +1201,9 @@ function Show-ADTInstallationWelcome
                             if ($adtSession)
                             {
                                 Update-ADTDeferHistory
-                                Close-ADTSession -ExitCode $adtConfig.UI.DefaultExitCode
+                                $sessionClosed = $true; Close-ADTSession -ExitCode $adtConfig.UI.DefaultExitCode
                             }
-                            else
-                            {
-                                return
-                            }
+                            return
                         }
                         elseif ($promptResult.Equals([PSADT.UserInterface.DialogResults.CloseAppsDialogResult]::Defer))
                         {
@@ -1223,12 +1221,9 @@ function Show-ADTInstallationWelcome
                             if ($adtSession)
                             {
                                 Update-ADTDeferHistory
-                                Close-ADTSession -ExitCode $adtConfig.UI.DeferExitCode
+                                $sessionClosed = $true; Close-ADTSession -ExitCode $adtConfig.UI.DeferExitCode
                             }
-                            else
-                            {
-                                return
-                            }
+                            return
                         }
                         elseif (!$promptResult.Equals('TerminatedTryAgain'))
                         {
@@ -1242,6 +1237,12 @@ function Show-ADTInstallationWelcome
                             }
                             throw (New-ADTErrorRecord @naerParams)
                         }
+                    }
+
+                    # Break if the session has closed as Close-ADTSession won't be able to break out of the above while loop.
+                    if ($sessionClosed)
+                    {
+                        break
                     }
 
                     # Close any remaining processes that are open that the user couldn't close.
--- a/src/PSAppDeployToolkit/Public/Start-ADTProcess.ps1
+++ b/src/PSAppDeployToolkit/Public/Start-ADTProcess.ps1
@@ -885,6 +885,7 @@ function Start-ADTProcess
             }
 
             # Switch on the exception type's name.
+            $sessionClosed = $false
             switch -Regex ($_.Exception.GetType().FullName)
             {
                 '^System\.Runtime\.InteropServices\.ExternalException$'
@@ -907,6 +908,7 @@ function Start-ADTProcess
                     Invoke-ADTFunctionErrorHandler @iafehParams -Silent
                     if ($iafehParams.ContainsKey('ErrorAction'))
                     {
+                        $sessionClosed = $true
                         Close-ADTSession
                     }
                     break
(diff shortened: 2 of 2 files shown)


DanGough:
There's a hard-coded user SID in https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/pull/1862/commits/bb493766409febcecced01cbba9cd2843286e4d7 ?

I've yet to do testing around #1838, I'll try and get to that tomorrow!

mjr4077au:
Good shout, genuine cockup. I've pushed a fix for that.

DanGough:
Current behaviour when there is a process running as system (e.g. Notepad) that's being requested to be closed:

- If you are an admin user, you see the process listed in the dialog, and you can close it (same as previous 4.1.x)
- If you are a standard user, the process is not listed in the dialog, but if you proceed, the toolkit will force close the process regardless from the main install context.

🐛 **There is a bug here however** - the new force close logic runs regardless which option the user selects, so _the process is force-closed even upon deferral or timeout_.

Once that issue is addressed - I'm not a fan of the fact that it only displays the system process to admin users, yet any user type can effectively close it. I would prefer that we either:

1. Only show the processes in the dialog that are running as the logged in user. Then std/admin users get the same experience, and the system process gets force-closed in the same way.
2. Display the all processes to all users, regardless if they are admins or not.

My preference is leaning towards `#1`.

DanGough:
As for #1841, where we've added a new `$RunAsActiveUserLocale` variable - we already set a variable `$uiculture`, that I would expect to match the locale of user who sees the UI. Could you please clarify why we could not just stick to that variable rather than adding a new one?

DanGough:
`Get-ADTUserProfiles` is returning empty 'UserLocale' values for me. Is this expected behaviour, i.e. you only want this field populated if the user has overridden the default language?

This puts more onus on the user to check if UserLocale is set and if not fall back to the system default - whereas if we did that ourselves and always ensured UserLocale contained something, this would save them a job.

mjr4077au:
> Get-ADTUserProfiles is returning empty 'UserLocale' values for me. Is this expected behaviour, i.e. you only want this field populated if the user has overridden the default language?

> This puts more onus on the user to check if UserLocale is set and if not fall back to the system default - whereas if we did that ourselves and always ensured UserLocale contained something, this would save them a job.

This is from a different computer to the one I developed on, but its reporting the UserLocale for me and I don't remember setting anything other than the regional settings during the OOBE.
```powershell
PS C:\Repos\Devicie-Repo-TelemetryPublisherWindows\Devicie\Devicie\bin\Debug\net472> Get-ADTUserProfiles -LoadProfilePaths


NTAccount              : DESKTOP-1UMKPSD\mjr40
SID                    : S-1-5-21-2555773579-1665067283-2784874702-1002
ProfilePath            : C:\Users\mjr40
AppDataPath            : C:\Users\mjr40\AppData\Roaming
LocalAppDataPath       : C:\Users\mjr40\AppData\Local
DesktopPath            : C:\Users\mjr40\OneDrive\Desktop
DocumentsPath          : C:\Users\mjr40\OneDrive\Documents
StartMenuPath          : C:\Users\mjr40\AppData\Roaming\Microsoft\Windows\Start Menu
TempPath               : C:\Users\mjr40\AppData\Local\Temp
OneDrivePath           : C:\Users\mjr40\OneDrive
OneDriveCommercialPath :
UserLocale             : en-AU

NTAccount              : DESKTOP-1UMKPSD\WsiAccount
SID                    : S-1-5-21-2555773579-1665067283-2784874702-1003
ProfilePath            : C:\Users\WsiAccount
AppDataPath            : C:\Users\WsiAccount\AppData\Roaming
LocalAppDataPath       : C:\Users\WsiAccount\AppData\Local
DesktopPath            : C:\Users\WsiAccount\Desktop
DocumentsPath          : C:\Users\WsiAccount\Documents
StartMenuPath          : C:\Users\WsiAccount\AppData\Roaming\Microsoft\Windows\Start Menu
TempPath               : C:\Users\WsiAccount\AppData\Local\Temp
OneDrivePath           :
OneDriveCommercialPath :
UserLocale             : en-AU

NTAccount              : DESKTOP-1UMKPSD\ExistingEndUser
SID                    : S-1-5-21-2555773579-1665067283-2784874702-1004
ProfilePath            : C:\Users\ExistingEndUser
AppDataPath            : C:\Users\ExistingEndUser\AppData\Roaming
LocalAppDataPath       : C:\Users\ExistingEndUser\AppData\Local
DesktopPath            : C:\Users\ExistingEndUser\Desktop
DocumentsPath          : C:\Users\ExistingEndUser\Documents
StartMenuPath          : C:\Users\ExistingEndUser\AppData\Roaming\Microsoft\Windows\Start Menu
TempPath               : C:\Users\ExistingEndUser\AppData\Local\Temp
OneDrivePath           : C:\Users\ExistingEndUser\OneDrive
OneDriveCommercialPath :
UserLocale             : en-AU

NTAccount              : Default
SID                    : S-1-0-0
ProfilePath            : C:\Users\Default
AppDataPath            : C:\Users\Default\AppData\Roaming
LocalAppDataPath       : C:\Users\Default\AppData\Local
DesktopPath            : C:\Users\Default\Desktop
DocumentsPath          : C:\Users\Default\Documents
StartMenuPath          : C:\Users\Default\AppData\Roaming\Microsoft\Windows\Start Menu
TempPath               : C:\Users\Default\AppData\Local\Temp
OneDrivePath           :
OneDriveCommercialPath :
UserLocale             :



PS C:\Repos\Devicie-Repo-TelemetryPublisherWindows\Devicie\Devicie\bin\Debug\net472>
```
If the value's truly not there though, I don't think it should pull some random value from somewhere else and insinuate its the caller's UserLocale though. The caller should interpret null however they wish and make a decision thereafter.

> As for https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1841, where we've added a new $RunAsActiveUserLocale variable - we already set a variable $uiculture, that I would expect to match the locale of user who sees the UI. Could you please clarify why we could not just stick to that variable rather than adding a new one?

The `$uiculture` variable's value has always been that of the caller (i.e. SYSTEM). Changes to that value could cause problems for users expecting it to be the calling account's locale. Further to this, it's not safe to do for multi-user environments because its somewhat pot luck who the active user is.

This needs to be part of a larger discussion beyond this PR because for me, the `$uilocale` is the least of our worries. The ones of higher concern are the things like `$envUserDesktop` as having these refer to SYSTEM locations is even more useless. I'd probably use that chat as a springboard for _reducing_ the variables we create though rather than fixing this up.

> Current behaviour when there is a process running as system (e.g. Notepad)

I'm reviewing this now and will come back later about it.

mjr4077au:
> `Get-ADTUserProfiles` is returning empty 'UserLocale' values for me. Is this expected behaviour, i.e. you only want this field populated if the user has overridden the default language?
> 
> This puts more onus on the user to check if UserLocale is set and if not fall back to the system default - whereas if we did that ourselves and always ensured UserLocale contained something, this would save them a job.

Superseding the last comment, pull again and see whether there's any fallback improvement for you.

DanGough:
My device does have the first reg key that you're looking for set:

HKEY_CURRENT_USER\Control Panel\International
LocaleName = en-GB

The issue was mine - resolved by specifying `Get-ADTUserProfiles -LoadProfilePaths`! This was originally just intended to load the registry hives to get extra paths back (e.g. Desktop). If it's needed for locale also, we'll either need to update the docs, or just remove the parameter and make it default... Initial concern though was that his might be expensive with lots of user profiles on board. Difference is negligible on my test machine with a single account.

mjr4077au:
> The issue was mine - resolved by specifying `Get-ADTUserProfiles -LoadProfilePaths`!

Ah, yeah, it can't just pull it out of thin air haha. The code here is more of a feature add than anything else, we don't use this code pathway in the module initialisation to retrieve the locale. You'll also see I've rewritten it to use Microsoft.Win32.Registry so it's a lot faster.

DanGough:
One last thing to discuss:

If the active user is an admin, they are shown and can close processes running as system. Should they instead just get the same result as a standard user, and only be presented with processes running as themselves?

With the current setup, a packager could put a system process or service in the list and have it close when they test it under an admin account, then it fails later when deploying to standard users where the system processes are ignored.

PR #1863, opened 2025-10-25
Fix return type for `Get-ADTUserNotificationState`.

mjr4077au:
Enum type didn't survive the serialisation round trip.


DanGough (commented):
Does this fix a specific issue, reported or not, for the benefit of the changelog?

mjr4077au:
Yeah, it's a fix for an issue.

Issue #1912, opened 2025-11-27
[Bug] Installing an MSI Patch (.msp) file using Start-MspProcess does not allow the creation of a Log file


AScott-WWF:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.7

### Describe the bug

I have been having fun and games trying to install a large MSP file to patch an even larger MSI install for ArcGIS Pro today.
My `Start-MspProcess` command line is as follows:
```
Start-ADTMspProcess -FilePath "$($mspPath.FullName)" -AdditionalArgumentList "ALLUSERS=1 REINSTALLMODE=omus REINSTALL=ALL /LOG='C:\ProgramData\Microsoft\IntuneManagementExtension\Logs\CompanyName-ArcGISPro_Patch_install.log' /NORESTART"
```
N.B. $mspPath is populated with this:
```
$files = Get-ChildItem -Path "$($adtSession.DirFiles)" -File -Recurse -ErrorAction SilentlyContinue
    $mspPath = $files | Where-Object { $_.Name -like 'ArcGIS_Pro_*.msp' }
```
When the Start-MspProcess executes it displays the following screen output:
```
[2025-11-27T16:58:14.2131343+00:00] [Install] [Start-ADTProcess] [Info] :: Executed ["C:\Windows\system32\msiexec.exe" /update "C:\Users\WDAGUtilityAccount\Downloads\ArcGIS Pro 3.5.5\Files\ArcGIS_Pro_355_195505.msp" REBOOT=ReallySuppress /QN REINSTALLMODE=ecmus REINSTALL=ALL ALLUSERS=1 REINSTALLMODE=omus REINSTALL=ALL /LOG='C:\ProgramData\Microsoft\IntuneManagementExtension\Logs\WWFUK-ArcGISPro_Patch_install.log' /NORESTART], awaiting completion...
```
and the Windows Installer (msiexec syntax) dialog is shown- this usually occurs if the syntax is incorrect.

[...]

PR #1916, opened 2025-11-29
Expose `Start-ADTMsiProcess`'s logging parameters through to `Start-ADTMspProcess`.

mjr4077au:
This is about the best we can do here as there's no reliable way to get the version out of an MSP file.

Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1912.

### DIFF
--- a/src/PSAppDeployToolkit/Public/Start-ADTMspProcess.ps1
+++ b/src/PSAppDeployToolkit/Public/Start-ADTMspProcess.ps1
@@ -21,6 +21,14 @@ function Start-ADTMspProcess
     .PARAMETER AdditionalArgumentList
         Additional parameters.
 
+    .PARAMETER LoggingOptions
+        Overrides the default logging options specified in the config.psd1 file.
+
+    .PARAMETER LogFileName
+        Overrides the default log file name. The default log file name is generated from the MSI file name. If LogFileName does not end in .log, it will be automatically appended.
+
+        For uninstallations, by default the product code is resolved to the DisplayName and version of the application.
+
     .PARAMETER RunAsActiveUser
         A RunAsActiveUser object to invoke the process as.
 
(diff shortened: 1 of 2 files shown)


DanGough:
This is how I get the version of an MSP, works for everything I've chucked at it so far:

```powershell
try {
    $WindowsInstaller = New-Object -ComObject WindowsInstaller.Installer
    [xml]$MSPPatchXML = $WindowsInstaller.ExtractPatchXMLData($Path)
    $Version = $MSPPatchXML.MsiPatch.TargetProduct.UpdatedVersion | Sort-Object { [version]$_ } -Bottom 1
}
catch {
    throw "Error querying MSP: $_"
}
finally {
    if ($WindowsInstaller) { [System.Runtime.InteropServices.Marshal]::ReleaseComObject($WindowsInstaller) | Out-Null }
}
```

In hindsight I would improve the Sort-Object piece, since that would fall over if it runs into a version represented by a single digit or some other invalid value.

DanGough (approved):
Approved, but leaving open in case you want to make use of the above to do something with the MSP version.

mjr4077au:
@DanGough nice one mate, ta for that. I wonder if it's always present and is mandatory though given https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1921 indicating it may not be?

DanGough:
I'd expect it to alway be present in an MSP patch since MSPs have metadata in stating which ProductCode and version ranges they're applicable to. I'm not sure if a versionless MSP would even be possible!

As long as we have some kind of fallback so that we ignore the version if its not found rather than bombing the script, we should be good.

mjr4077au:
@DanGough I'd make sure there's fallback for sure. I'll implement this in a separate PR as I'll do it via a P/Invoke to avoid all the COM nonsense (https://learn.microsoft.com/en-us/windows/win32/api/msi/nf-msi-msiextractpatchxmldataw).

PR #1935, opened 2025-12-11
Parallelize code signing

pl4nty:
Code signing in CI currently takes ~9 minutes, likely because it's signing each file synchronously. Let's use AzureSignTool's native parallelism to try and speed it up

Untested, but I've used a similar pattern elsewhere


pl4nty:
image

Issue #1925, opened 2025-12-09
[Feature] Unseal PSADT.Types.InstalledApplication


neo42JBR:
### Summary of the new feature / enhancement

In the light of extension development I'd be happy if the InstalledApplication type would be unsealed to allow inheritance. 

This way a extension module developer can add additional information to applications.
It can be useful for more advanced filtering or providing different kinds of applications (e.g. AppX)

Edit:
Alternatively add an overflow dictionary that contains the keys that do not belong to the static definition.
One could use that to store the extra information when creating new instances

mjr4077au:
> It can be useful for more advanced filtering or providing different kinds of applications (e.g. AppX)

In light of the above and with https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1450 pending also, is there the possibility to collaborate on a mutual AppX solution involving this data type?

tjgruber:
> > It can be useful for more advanced filtering or providing different kinds of applications (e.g. AppX)
> 
> In light of the above and with [#1450](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1450) pending also, is there the possibility to collaborate on a mutual AppX solution involving this data type?

I implemented Appx handling in a kind of hacky way that uses and extends the `PSADT.Types.InstalledApplication` properties for proper downstream handling, such as enabling `Uninstall-ADTApplication` to function with Appx apps (supports msix, bundle, etc.).

It's a bit messy, but didn't need to spend much time on it, and has been working well managing a number of Appx type apps at scale.

**Extension here:** https://github.com/tjgruber/Intune/blob/main/PSADT%20Extensions/Invoke-MDMAOAppxAppDetectionExt.ps1

I'm using it with this calling function: https://github.com/tjgruber/Intune/blob/main/PSADT%20Extensions/Invoke-MDMAOClassicAppDetectionExt.ps1

Here's an example detection output:

<img width="2061" height="952" alt="Image" src="https://github.com/user-attachments/assets/161c534a-7318-4193-9a74-64427e93654b" />

Uninstall example:

<img width="2052" height="279" alt="Image" src="https://github.com/user-attachments/assets/5471c751-a482-4847-9088-9d6b76b44426" />

mjr4077au:
Closing this one so we can just focus on https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1450.

PR #1936, opened 2025-12-11
Add Appx capabilties 

neo42JBR:
The implementation shown provides the c# backend to work with Appx packages.

A demo implementation was added to `Get-ADTApplication` to preview its capabilities.
The function implementation is limited to provisioned packages, as this reflects "system installed" packages equivalent to how ARP entries are queried.

The backend exposes methods to work with both provisioned and installed packages, either per user or any.
All methods are purely static and work solely on an identifier basis.
I avoided adding requirements for the Windows SDK to handle appx management to reduce external dependencies.
Information about packages is extracted from the registry and the manifest files they point to.

Filtering for family names is supported, however currently there is no property to expose this property in `InstalledApplication`.

### DIFF
--- a/src/PSADT/PSADT/Types/InstalledApplication.cs
+++ b/src/PSADT/PSADT/Types/InstalledApplication.cs
@@ -15,6 +15,18 @@ namespace PSADT.Types
         /// <summary>
         /// Initializes a new instance of the <see cref="InstalledApplication"/> record.
         /// </summary>
+        /// <param name="psPath">The registry key that contains the uninstall entry.</param>
+        /// <param name="psParentPath">The registry key for the subkey's parent.</param>
+        /// <param name="psChildName">The registry subkey for uninstalling the application.</param>
+        /// <param name="displayName">The display name of the application.</param>
+        /// <param name="displayVersion">The version of the application.</param>
+        /// <param name="uninstallString">The uninstall string used to remove the application.</param>
+        /// <param name="installSource">The source from which the application was installed.</param>
+        /// <param name="installLocation">The location where the application is installed.</param>
+        /// <param name="installDate">The date the application was installed (as a string).</param>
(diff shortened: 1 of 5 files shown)


neo42JBR:
Technically the ability to read the manifest information from an installer would be easily implemented as the bundle can be read as zip and the manifest file could be extracted.
However importing System.IO.Compression collides with .NET 8.

DanGough:
Sorry, Next was a temporary branch, this PR was not closed intentionally. @mjr4077au will be in touch!

neo42JBR:
After discussion with @mjr4077au this will require #1925.
Most of the code will also be ported to PS where applicable

neo42JBR:
The commit I just pushed already implements #1925, without full wiring in the toolkit yet.
The structure of inheritance is the following:
```
                       | -> InstalledArpApplication -> InstalledMsiApplication
InstalledApplication - |
                       | -> InstalledAppxApplication
```

I also dropped everything except the `GetPackageFamilyName` function from c# to powershell.

Nearly all logic is now packed into the new `Get-ADTPackageManifest` function that can read package info from an installer (like we do with MSI) or any other valid manifest file.

Sorry @mjr4077au, I had to force push because most of the changes you applied were not relevant anymore. The ones that still were I tried to apply where applicable.

JonathanPitre:
I'm so happy you guys are making progress on this. I'm using custom posh code to get this working and it's far from easy. 

Slack and 1Password nows recommends MSIX  deployment and other manufacturers will follow soon. The addition of AppX support is a must and PSADT is leading the way. 😁

neo42JBR:
This is only a draft PR. During development it became apparent that this requires more changes than originally expected.
This is probably best splitting into multiple seperate PRs, however I am currently not able to work on this. Just leaving this up for now as future reference.

PR #1944, opened 2025-12-17
Ensure compiled .psm1 is UTF8 with BOM

DanGough:
Ensure compiled .psm1 is UTF8 with BOM.

### DIFF
--- a/src/PSAppDeployToolkit.build.ps1
+++ b/src/PSAppDeployToolkit.build.ps1
@@ -780,7 +780,7 @@ Add-BuildTask Build {
         # Write out the processed file back to disk.
         $text; [System.String]::Empty; [System.String]::Empty
     }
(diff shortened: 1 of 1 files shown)


src/PSAppDeployToolkit.build.ps1
```diff
@@ -780,7 +780,7 @@ Add-BuildTask Build {
         # Write out the processed file back to disk.
         $text; [System.String]::Empty; [System.String]::Empty
     }
-    [System.IO.File]::WriteAllLines($Script:BuildModuleRootFile, $scriptContent)
+    [System.IO.File]::WriteAllLines($Script:BuildModuleRootFile, $scriptContent, [System.Text.Encoding]::UTF8)
```
mjr4077au: This should be `[System.Text.UTF8Encoding]::new($true, $true)` instead as the default encoding for UTF8 is to not include a BOM.

mjr4077au (changes requested):

mjr4077au (approved):

PR #1968, opened 2026-01-14
Remove `Start-ADTMs*ProcessAsUser` from the toolkit.

mjr4077au:
# Pull Request

## Description

These were added for parity with `Start-ADTProcessAsUser`, but they just support bad deployment habits: https://discourse.psappdeploytoolkit.com/t/install-as-logged-on-user-for-a-msi/7252. Adding these were a mistake on my part and I regret that I bought them into the toolkit.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


DanGough:
Agreed that using this function to install stuff in user context from the system account is generally dumb, and doesn't magically fix user-context installs that need elevation like some might hope. But I see that as a documentation/education issue - it is already mentioned in the docs in at least one place but could be repeated wherever it's relevant.

I have encountered the odd app where you're installing a bunch of components and one of them happens to be user context, and being able to slip in a per-user install amongst the rest without having to make it a separate app/dependency in Intune can be pretty handy.

However the main use I have for this function, something we have even got PMPC customers using successully to resolve issues, is to use it to **remove** user-context installs when deploying the system-wide version of an app. If these functions are removed, it would break their scripts and force them to jump through extra hoops to use `Start-ADTProcessAsUser` instead.

Therefore I vote to keep it in and improve the docs, maybe even print a warning in the log if people use it with `-Action Install`.

mjr4077au:
How about making them private then?

mjr4077au:
As discussed privately, let's not as there's legitimate uses as mentioned above by @DanGough. I suppose at some point, if users want to do silly things then we shouldn't gimp functionality to prevent the silliness at the expense of legitimate use cases.

Issue #1973, opened 2026-01-16
[Bug] Zero-Config MSI installation is failing


johndoe27:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.8

### Describe the bug

Error
```
[2026-01-16T10:21:26.3642540-05:00] [Initialization] [Open-ADTSession] [Info] :: -------------------------------------------------------------------------------
[2026-01-16T10:21:26.3955412-05:00] [Initialization] [Open-ADTSession] [Info] :: Discovered Zero-Config MSI installation file [C:\Users\WDAGUtilityAccount\Desktop\Tmp\Packages\Google_Chrome_v144.0.7599.60_REV1 - New\Toolkit\Files\googlechromestandaloneenterprise64.msi].
[2026-01-16T10:21:26.6146611-05:00] [Initialization] [Open-ADTSession] [Error] :: Failure occurred while instantiating new deployment session: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at PSADT.Module.DeploymentSession..ctor(IReadOnlyDictionary`2 parameters, Nullable`1 noExitOnClose, SessionState callerSessionState) in D:\a\PSAppDeployToolkit\PSAppDeployToolkit\src\PSADT\PSADT\Module\DeploymentSession.cs:line 357
[2026-01-16T10:21:26.6307423-05:00] [Initialization] [Open-ADTSession] [Error] :: PSAppDeployToolkit deployment completed in [0.2664883] seconds with exit code [60008].
[2026-01-16T10:21:26.6307423-05:00] [Initialization] [Open-ADTSession] [Info] :: -------------------------------------------------------------------------------
Open-ADTSession : Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
[...]

Issue #1971, opened 2026-01-16
[Bug] Copy-ADTFile (folder) places the folder inside a folder of the same name (doco update)


Random206:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.7

### Describe the bug

When using Copy-ADTFile and you copy a folder recursively the documentation found [here](https://psappdeploytoolkit.com/docs/reference/functions/Copy-ADTFile) in example 2, says 

> Copy-ADTFile -Path 'C:\Path\Folder' -Destination 'D:\Destination\Folder' -Recurse
> Recursively copies the folder 'Folder' from 'C:\Path' to 'D:\Destination'.

However doing this creates D:\Destination\Folder\Folder\\{Contents}

The documentation should show: `Copy-ADTFile -Path 'C:\Path\Folder' -Destination 'D:\Destination' -Recurse`

Perhaps the doco needs to be less ambiguous and use terms like C:\Path\FolderA for the source and D:\Destination\FolderB if you want to put FolderA and it's contents inside FolderB.

### Steps to reproduce

1. Copy a folder using -recursive and it will include the source folder and put it into the destination folder.
2. Example: Copy-ADTFile -Path 'C:\Path\Folder' -Destination 'D:\Destination\Folder' -Recurse

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 24H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 26100.1.amd64fre.ge_release.240331-1435
OsLanguage           : en-GB
OsMuiLanguages       : {en-GB, en-US}
KeyboardLayout       : en-US
TimeZone             : (UTC+10:00) Brisbane
HyperVisorPresent    : True
CsPartOfDomain       : True
CsPCSystemType       : Desktop


Host:
  Version:      8.0.21
  Architecture: x64
  Commit:       362ab6669d
  RID:          win-x64

[...]

Issue #1969, opened 2026-01-14
[Bug] CloseProcesses not working - LibreOffice


gunwee:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.5

### Describe the bug

CloseProcesses does not seem to work for LibreOffice.
It's a bit messy. Each program start their own process (scalc,swriter,smath,simpress,sdraw) until welcome message is gone. Then they merge to soffice.exe and/or soffice.bin
I get all kinds of different outcomes depending on which process is started. But none ever gets successfully closed, even if the log says so.
Seem to work with _kinda_ work with **-PromptToSave** but still wonky.

[LibreOffice25.8.4.2_25.8.4.2_EN_01_PSAppDeployToolkit_Uninstall.log](https://github.com/user-attachments/files/24615220/LibreOffice25.8.4.2_25.8.4.2_EN_01_PSAppDeployToolkit_Uninstall.log)

### Steps to reproduce

ZeroConfigMSI.

```
$adtSession = @{
    # App variables.
    AppVendor = ''
    AppName = ''
    AppVersion = ''
    AppArch = ''
    AppLang = 'EN'
    AppRevision = '01'
    AppSuccessExitCodes = @(0)
    AppRebootExitCodes = @(1641, 3010)
    AppProcessesToClose = @('sbase','scalc','sdraw','senddoc','simpress','smath','soffice','soffice.bin','soffice_safe','spsupp_helper','sweb','swriter')  # Example: @('excel', @{ Name = 'winword'; Description = 'Microsoft Word' })
    AppScriptVersion = '1.0.0'
    AppScriptDate = '2025-08-21'
    AppScriptAuthor = ''
    RequireAdmin = $true
    
    # Install Titles (Only set here to override defaults set by the toolkit).
    InstallName = ''
    InstallTitle = ''

    # Script variables.
    DeployAppScriptFriendlyName = $MyInvocation.MyCommand.Name
    DeployAppScriptParameters = $PSBoundParameters
    DeployAppScriptVersion = '4.1.3'
        
    ForceWimDetection = $true
[...]

Issue #1966, opened 2026-01-12
[Bug] Processes not closed / blocked, but setup starts running


alexhass:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.7

### Describe the bug

We identified an issue today when updating VSCode. The developers had 4 x "code.exe" open. PSADT started to complain that the EXE need to be closed. After the timeout it closed ONE code.exe. The other 3 applications stayed open, flickered in red to safe data  - but user was away from desk . Applications have not killed.

I guess PSADT is not closing all applications, but just one?

This has than caused a setup failure with error code 5 (what is not correct - the user was on lunch):
- 5	The user clicked Cancel during the actual installation process, or chose Abort at an Abort-Retry-Ignore box.

Result: Installation was broken and lots of files missing on disk.

Here are my saiwParams:
```
	$saiwParams = @{
		AllowDeferCloseProcesses = $true
		AllowMove = $true
		BlockExecution = $true
        CheckDiskSpace = $true
		DeferRunInterval = ([System.TimeSpan]::FromHours(1))
        DeferTimes = 3
		ForceCloseProcessesCountdown = 600
        PersistPrompt = $true
		PromptToSave = $true
    }
```

I'm also attaching the full logs of both PSADT and the InnoSetup for better understanding.

[REDACTED_Microsoft Visual Studio Code_1.107.1_Install.log](https://github.com/user-attachments/files/24569965/REDACTED_Microsoft.Visual.Studio.Code_1.107.1_Install.log)
[REDACTED_MicrosoftVisualStudioCode_1.107.1_x64_EN_01_PSAppDeployToolkit_Install.log](https://github.com/user-attachments/files/24569964/REDACTED_MicrosoftVisualStudioCode_1.107.1_x64_EN_01_PSAppDeployToolkit_Install.log)

### Steps to reproduce

1. Open 4 x VSCode applications
2. Run the PSADT install via Intune
3. First application process closes after timeout of 10 minutes
4. Other 3 process stay open and processes are not killed after timeout
[...]

PR #1975, opened 2026-01-17
Latest development changes sync to main.

mjr4077au:
# Pull Request

## Description

Latest development improvements, along with fixes for three outstanding issues.

- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/341
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1476
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1548
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1562
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1592
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1730
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1747
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1758
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1815
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1816
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1829
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1893
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1926
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1966
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1969
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1971
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1973
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1979
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1980
- Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1983

## Type of change

## Checklist

### DIFF
--- a/src/PSAppDeployToolkit/Public/Install-ADTMSUpdates.ps1
+++ b/src/PSAppDeployToolkit/Public/Install-ADTMSUpdates.ps1
@@ -50,7 +50,7 @@ function Install-ADTMSUpdates
     (
         [Parameter(Mandatory = $true)]
         [ValidateScript({
-                if (!(Test-Path -LiteralPath $_ -PathType Container))
+                if (!(Test-Path -LiteralPath $_))
                 {
                     $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName LiteralPath -ProvidedValue $_ -ExceptionMessage 'The specified directory does not exist.'))
                 }
@@ -64,16 +64,23 @@ function Install-ADTMSUpdates
     {
         # Initialize function.
         Initialize-ADTFunction -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
-        if (!($updates = Get-ChildItem -Path "$LiteralPath\*.msu" -Recurse -ErrorAction Ignore))
+        if (!($updates = if ([System.IO.Directory]::Exists($LiteralPath)) { Get-ChildItem @PSBoundParameters -Filter *.msu -Recurse -ErrorAction Ignore } else { $LiteralPath }))
         {
-            $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName LiteralPath -ProvidedValue $_ -ExceptionMessage 'The specified directory contains no updates.'))
+            $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName LiteralPath -ProvidedValue $_ -ExceptionMessage 'The specified path contains no updates.'))
         }
     }
 
     process
     {
         # Get all hotfixes and install if required.
-        Write-ADTLogEntry -Message "Recursively installing all Microsoft Updates in directory [$LiteralPath]."
+        if ($updates -isnot [System.String])
+        {
+            Write-ADTLogEntry -Message "Recursively installing all Microsoft Updates in directory [$LiteralPath]."
+        }
+        else
+        {
+            Write-ADTLogEntry -Message "Installing Microsoft Update [$updates]."
+        }
         foreach ($update in $updates)
         {
             if ($PSCmdlet.ShouldProcess("Microsoft Update [$($update.Name)]", 'Install'))


src/PSAppDeployToolkit/Public/Uninstall-ADTApplication.ps1
```diff
@@ -354,7 +354,7 @@ function Uninstall-ADTApplication
                     Write-ADTLogEntry -Message "Removing EXE application [$($removeApplication.DisplayName)$(if ($removeApplication.DisplayVersion -and !$removeApplication.DisplayName.Contains($removeApplication.DisplayVersion)) { " $($removeApplication.DisplayVersion)" })]."
                     try
                     {
-                        Start-ADTProcess @sapParams -ErrorAction $OriginalErrorAction
+                        Start-ADTProcess @sapParams -CreateNoWindow:($uninstallProperty.Equals('QuietUninstallString')) -ErrorAction $OriginalErrorAction
```
DanGough: What's the reasoning behind this one?

Sometimes an app will only have an UninstallString, but the additional args would then usually be added via -AdditionalArgumentList - so I don't see why CreateNoWindow would change behaviour based on the default command line found in the registry.

src/PSAppDeployToolkit/Public/Remove-ADTDesktopShortcut.ps1
```diff
@@ -0,0 +1,202 @@
...
+            $PSCmdlet.ThrowTerminatingError($_)
+        }
+
+        # Establish lookup table for the differing scopes.
+        $adtEnv = Get-ADTEnvironmentTable
+        $scopeTable = @{
+            RunAsActiveUser = @{
+                Path = if ($adtEnv.RunAsActiveUser)
+                {
+                    (Get-ADTUserProfiles -SID $adtEnv.RunAsActiveUser.SID -LoadProfilePaths -InformationAction SilentlyContinue).DesktopPath.FullName
```
DanGough, line 98: Is there a possibility when running a user context deployment on a multi-user system that RunAsActiveUser could be somebody else who logged on after you? If so we might want a CurrentUser option that simply makes use of $envUserDesktop.

mjr4077au: If you look at line 96, we're using `$adtEnv.RunAsActiveUser`, so it's the RunAsActiveUser as established during module initialisation, which is consistent with the rest of the toolkit when we're doing something user-facing.

That `$envUserDesktop` from SYSTEM will be the SYSTEM profile's Desktop directory, so that's not reliable. We'll have to talk about that one in https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1902, but https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1536 will need to come before that.

src/PSAppDeployToolkit/Public/Uninstall-ADTApplication.ps1
mjr4077au: This one is tied to https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1816, which from the Discord chat at the time was that the caller was expecting to see the uninstall window as it was interactive, but it wasn't. The process code currently will hide any window if you specify `-CreateNoWindow`, so this was an attempt to alleviate that instead of blanketly enforcing it in `$sapParams`.

Given your (quite valid) feedback, maybe we just don't apply this switch at all? We don't enforce `-WindowStyle`, etc, so we probably shouldn't put `-CreateNoWindow` in either.

src/PSAppDeployToolkit/Public/Remove-ADTDesktopShortcut.ps1
mjr4077au: Oh, I see what you're asking. The answer's no, the function to determine the RunAsActiveUser during operations is rigged to use you at all if you're a validly logged on user (i.e. not a service account, or some other non-interactive account).

src/PSAppDeployToolkit/Public/Uninstall-ADTApplication.ps1
DanGough: Adding CreateNoWindow is handy for capturing stdout if it happens to be a console app... but hiding the UI can be a pain when you're trying to debug an uninstall that's hanging on you due to popping up an unsuppressed dialog.

Either way, we shouldn't be switching visibility based upon the presence of QuietUninstallString.

Is the idea behind #1816 then to try and determine if it's a console or GUI app before launching it?

mjr4077au: Agreed, it wasn't the right decision here. We have the means to detect whether the app is a console app or not, so I'll rework this shortly.

mjr4077au: @DanGough amended via https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/pull/1975/changes/a784c2850a280809782c01238bf6e4327139e24f.

Issue #1594, opened 2025-08-11
[Bug] Copy-ADTFile / Robocopy cannot handle literal file names with wildcard characters


kodemum:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1-rc3

### Describe the bug

I'm having an issue with copying a file using `Copy-ADTFile` with a name that can be resolved as path wildcard characters.
When `FileCopyMode` is set to use Robocopy, there doesn't seem to be any way to get this to work. 

The file name I am using for testing is `hi[].msi`

Wildcard characters not escaped

If the wildcard characters are not escaped, that is the command is as follows:

`Copy-ADTFile -Path "hi[].msi" -Destination "$envProgramFiles"`

Then the script fails with a **The specified wildcard character pattern is not valid** error.

[Robocopy_NotEscaped.log](https://github.com/user-attachments/files/21715117/Robocopy_NotEscaped.log)

However, if the characters are escaped:

```Copy-ADTFile -Path "hi ``[``].msi" -Destination "$envProgramFiles"```

Then the file is not copied.

[Robocopy_Escaped.log](https://github.com/user-attachments/files/21715118/Robocopy_Escaped.log)

I would think the best fix is probably to add a -LiteralPath property to Copy-ADTFile to be used when you do not want to use wildcards.

### Steps to reproduce

1. Set `FileCopyMode` to `Robocopy`
2. Attempt to copy a file with a name that contains a path wildcard (for example: `[]`)

### Environment data

```console
OsName               : Microsoft Windows 10 Enterprise N LTSC
OsOperatingSystemSKU : 126
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise N LTSC 2021
WindowsBuildLabEx    : 19041.1.amd64fre.vb_release.191206-1406
OsLanguage           : en-GB
OsMuiLanguages       : {en-GB}
KeyboardLayout       : 00020409
TimeZone             : (UTC-08:00) Pacific Time (US & Canada)
HyperVisorPresent    : True
CsPartOfDomain       : False
[...]

PR #1987, opened 2026-01-26
Add `-LiteralPath` support to `Copy-ADTFile`.

mjr4077au:
# Pull Request

## Description

Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1594.

## Type of change

## Checklist

### DIFF
--- a/src/Tests/Unit/Copy-ADTFile.Tests.ps1
+++ b/src/Tests/Unit/Copy-ADTFile.Tests.ps1
@@ -72,6 +72,12 @@ Describe 'Copy-ADTFile'-ForEach @(
             "$DestinationPath\test.txt" | Should -Exist
         }
 
+        It 'Copies a single file via -LiteralPath($PreCreateDestination = $<PreCreateDestination>; $FileCopyMode = $<FileCopyMode>)' {
+            Copy-ADTFile -LiteralPath "$SourcePath\test.txt" -Destination $DestinationPath -FileCopyMode $FileCopyMode
+
+            "$DestinationPath\test.txt" | Should -Exist
+        }
+
         It 'Copies a single file with a new filename ($PreCreateDestination = $<PreCreateDestination>; $FileCopyMode = $<FileCopyMode>)' {
             Copy-ADTFile -Path "$SourcePath\test.txt" -Destination "$DestinationPath\new.txt" -FileCopyMode $FileCopyMode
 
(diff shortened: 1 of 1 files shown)


DanGough:
I'd like to see the Pester test for this function extended to cover this new param before merging. Happy to take that on myself.

mjr4077au:
> I'd like to see the Pester test for this function extended to cover this new param before merging. Happy to take that on myself.

Thanks mate, that'd be much appreciated! I didn't mean to not amend them, just trying to churn and burn through the open issues but they rightly should be extended to test the `-LiteralPath` pathway.

DanGough:
It'd make sense to uplift Copy-ADTFileToUserProfiles also here.

src/PSAppDeployToolkit/Public/Copy-ADTFile.ps1
```diff
@@ -196,13 +204,14 @@ function Copy-ADTFile
...
                         # Trim ending backslash from paths which can cause problems with Robocopy.
                         # Resolve paths in case relative paths beggining with .\, ..\, or \ are used.
                         # Strip Microsoft.PowerShell.Core\FileSystem:: from the beginning of the resulting string, since Resolve-Path adds this to UNC paths.
-                        $robocopySource = (Get-Item -Path $srcPath.TrimEnd('\') -Force).FullName -replace '^Microsoft\.PowerShell\.Core\\FileSystem::'
-                        $robocopyDestination = (Join-Path -Path ((Get-Item -LiteralPath $Destination -Force).FullName -replace '^Microsoft\.PowerShell\.Core\\FileSystem::') -ChildPath (Split-Path -Path $srcPath -Leaf)).Trim()
+                        $getItemSplat = @{ $PSCmdlet.ParameterSetName = $srcPath.TrimEnd('\') }
+                        $robocopySource = (Get-Item @getItemSplat -Force).FullName -replace '^Microsoft\.PowerShell\.Core\\FileSystem::'
+                        $robocopyDestination = (Join-Path -Path ((Get-Item -LiteralPath $Destination -Force).FullName -replace '^Microsoft\.PowerShell\.Core\\FileSystem::') -ChildPath (Split-Path @pathSplat -Leaf)).Trim()
```
DanGough: `Split-Path -LiteralPath` does not support `-Leaf` in its parameter set:

Image

Destination is always a literal path, but we're appending the leaf of the source path which could contain a wildcard... So I can foresee this having an issue though when using `-Path "C:\Program Files*" in Robocopy mode, as it will try to use * in the destination folder name also. Announce Robocopy deprecation in 4.3.0 and move to its own function?!

mjr4077au:
> It'd make sense to uplift Copy-ADTFileToUserProfiles also here.

This is done and pushed.

src/PSAppDeployToolkit/Public/Copy-ADTFile.ps1
mjr4077au: The sooner we shitcan the Robocopy stuff, the better. Good shout on the `Split-Path` stuff as well. This is where PSScriptAnalyzer is an incredible disappointment. You could never have such a mistake slip through in C#, etc.

mjr4077au: @DanGough this is also done and pushed. Interesting the unit tests didn't pick any of this up though. Maybe they're not as comprehensive as they could be?

DanGough: I didn't even realise Split-Path supported both so thought I'd try it in case it ever pumped out an array when giving a wildcard to match... But the whole function seems pretty crippled with -LiteralPath, only giving you the left hand side of the path for god knows what reason.

mjr4077au: Yeah 100% mate, I clearly didn't realise either! As said it was a good spot, hopefully things are more amenable now for your review.

Issue #1414, opened 2025-04-13
[Feature] Write unit/integration tests for everything to improve toolkit reliability


mjr4077au:
### Summary of the new feature / enhancement

Some help for this would be great, there's a lot of PowerShell functions and C# code to cover off.

PR #2062, opened 2026-03-18
Add Pester tests for `Test-ADTRegistryValue`

michael-hollingsworth:
# Pull Request

## Description

Part of: <https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1414>

## Type of change

## Checklist

## How Has This Been Tested?

```PowerShell
. .\build.ps1
```

### DIFF
--- a/src/Tests/Unit/Test-ADTRegistryValue.Tests.ps1
+++ b/src/Tests/Unit/Test-ADTRegistryValue.Tests.ps1
@@ -14,10 +14,10 @@ Describe 'Test-ADTRegistryValue' {
 
     Context 'Functionality' {
         It 'Should return $true' {
-            Test-ADTRegistryValue -Key $TestRegistry -Name 'Test' | Should -Be $true
+            Test-ADTRegistryValue -Key $TestRegistry -Name 'Test' | Should -BeTrue
         }
         It 'Should return $false' {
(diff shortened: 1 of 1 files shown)


michael-hollingsworth:
By the way, I found that when using `TestRegistry:\TestLocation` as the value for the `-Key` parameter, an error would be thrown due to `Convert-ADTRegistryPath` not being able to determine the registry hive.

mjr4077au:
Hey there, thanks for working on this, it's very much appreciated! 

> By the way, I found that when using `TestRegistry:\TestLocation` as the value for the `-Key` parameter, an error would be thrown due to `Convert-ADTRegistryPath` not being able to determine the registry hive.

For now, this is probably acceptable as it's clearly a path that wouldn't convert, unless someone went out of their way to set up a PSDrive explicitly named that.

Ideally we'd use something like `$PSCmdlet.GetUnresolvedProviderPathFromPSPath()` instead of our own regex fluff, but I had issues with this when I tried last year. You can see I made such a change via ad7dbc867a296be91cf26aa1ac7d45412370d740, then reverted it via a739bf88837b551a4d263ba0c97974950729118e. I can't recall exactly what went on there and my commit notes are a bit terse unfortunately.

Is there somewhere we can chat at all? Perhaps on Discord via DM or something? It'd be nice to introduce myself properly and give you a proper thank you for your work. My handle over there is the same as here. You can also find me in the WinAdmins server as well.

PR #2069, opened 2026-03-20
Refactor `ProcessManager.LaunchAsync()` to remove all the CA2000 suppressions.

mjr4077au:
# Pull Request

## Description

The `ProcessManager.LaunchAsync()` method is one of our most complex, and important blocks of code within the repository. As it returns a handle containing a Task that can be awaited on, several disposable objects on the stack of `ProcessManager.LaunchAsync()` needed to survive once `ProcessManager.LaunchAsync()` returns said handle. The compiler could never quite understand what we were doing so the warnings were suppressed as they were benign.

This PR refactors `ProcessManager.LaunchAsync()` by moving certain elements of code into a new state class object that is responsible for ownership of a number of handles at various steps along the way. This refactor removes the CA2000 elements as the compiler has a better understanding of what's going on.

In addition to this, we have a number of other improvements, such as:

* Better object disposal if there's an exception thrown along the way.
* Remove all `STARTUPINFOW` copying as it can be passed as an `in` parameter.
* Proper async support for the non-job code pathway by way of some polyfills.
* No longer forcing `-Timeout` code through the job pathway as it's not necessary now.
* Removed some other memory copy operations that weren't necessary.
* Better stdin handling if the process doesn't read all of stdin before exiting.
* Address all resolvable CodeAnalysis suppressions within the codebase.

## Type of change

## Checklist

## How Has This Been Tested?

* Built module through to completion, which runs Robocopy.exe via `Copy-ADTFile` unit tests.
* Tested manual process creation with and without timeouts.
* Tested manual process creation with and without job features.
* Tested manual process creation with and without UseShellExecute.
* Tested toolkit interactively to validate our client/server architecture.


.github/copilot-instructions.md
```diff
@@ -51,3 +54,17 @@ The main build dependency is a current .NET SDK. Visual Studio is recommended fo
 - `PSADT.Interop` contains Win32 interop and CsWin32-generated symbols.
 - `PSADT` contains core C# utilities.
 - `PSAppDeployToolkit` contains PowerShell-facing C# types.
+- Apply `WIN32_LEAN_AND_MEAN` by default where appropriate to minimize native headers without prompting.
```
DanGough: Would these C# specific code guidelines be better off in the csharp.md file?

.github/copilot-instructions.md
```diff
@@ -51,3 +54,17 @@ The main build dependency is a current .NET SDK. Visual Studio is recommended fo
...
+- Use the existing Lock polyfill instead of object-lock suppression.
+- Prefer LoadLibraryEx over LoadLibrary for native loading.
+- Avoid direct P/Invoke when CsWin32 wrappers exist.
+- Initialize PSADT.Native loading in the NativeMethods static constructor.
+- Do not use `Condition="Exists(...)"` to silently skip required PSADT.Native asset copies; prefer fail-fast behavior that surfaces missing native DLLs as explicit build errors.
+
+## C++ Notes
+
+- C++ projects may utilize C++23 preview features. Use static lambdas for thread delegate declarations where applicable.
```
DanGough: cpp.md is referenced above but does not exist - this line could be a good starting point for that

.github/copilot-instructions.md
mjr4077au: It's where Copilot put the lines as it was making amendments, none of that's by hand. Maybe the splitting out into individual language instruction files isn't working right?

.github/copilot-instructions.md
mjr4077au: Again, it's another Copilit special as it's done this on its own accord. What's the expectation here with these instruction files? Is it meant to be able to manage itself?

src/PSADT/PSADT/ProcessManagement/ProcessLaunchInfo.cs
```diff
@@ -72,6 +72,12 @@ public ProcessLaunchInfo(string filePath, IEnumerable<string>? argumentList = nu
                 Verb = verb;
             }
 
+            // Confirm we're not using incompatible options.
+            if (useShellExecute && !(runAsActiveUser is null || runAsActiveUser == AccountUtilities.CallerRunAsActiveUser))
+            {
+                throw new NotSupportedException("Cannot specify UseShellExecute while specifying a RunAsActiveUser.");
```
DanGough, line 78: I can no longer run this, support for which was recently added and works in main:

```powershell
Start-ADTProcessAsUser -FilePath 'https://psappdeploytoolkit.com' -UseShellExecute
```
```
PSADT.ClientServer.ServerException: The client process returned an exception. ---> System.NotSupportedException: Running as a different user is not supported with ShellExecuteEx.
```

Similar error appears if trying to run the toolkit in user mode and run Start-ADTProcess:

```powershell
Start-ADTProcess -FilePath 'https://psappdeploytoolkit.com' -UseShellExecute
```
```
Start-ADTProcess : Exception calling "LaunchAsync" with "1" argument(s): "Running as a different user is not supported with ShellExecuteEx."
```

mjr4077au: Please pull down the latest, I put an extra commit in to fix this ~24 minutes ago.

src/PSAppDeployToolkit/Public/Start-ADTProcess.ps1
```diff
@@ -863,7 +863,7 @@ function Start-ADTProcess
                             Exception = [System.InvalidProgramException]::new("The launching of the process returned a null result.")
                             Category = [System.Management.Automation.ErrorCategory]::InvalidResult
                             ErrorId = 'ProcessExecutionNullReturn'
-                            TargetObject = $execution
+                            TargetObject = $(if (!$SecureArgumentList) { $execution })
```
DanGough, line 866: FYI there is still scope for the secret args to leak into the log via PositionMessage if the command line is sufficiently short:

Image

Typically this won't be a problem, I had to rename my MSI and trim down the command to get this.

mjr4077au: Probably not much we can do there without making Resolve-ADTErrorRecord not print out the PositionMessage, which almost feels not worth it.

DanGough: Yeah not expecting us to accomodate for this, just pointing it out. Property is read only but we can overwrite InvocationInfo.PositionMessage via Add-Member -Force if we wanted a hacky way to strip the data from the exception.

PR #2071, opened 2026-03-20
Add -Exclude parameter to Copy-ADTContentToCache. (#1740)

### LINKED ISSUE #1740: [Feature] -Skip 'Folder' on Copy-ADTContentToCache .

A -Skip Argument to Copy-ADTContentToCache 
like Copy-ADTContentToCache -LiteralPath "$env:TEMP\PSADT" -Skip 'SupportFiles','Files'

_No response_

DanGough:
# Pull Request

## Description

Adds -Skip parameter to Copy-ADTContentToCache, resolves #1740.

## Type of change

## Checklist

## How Has This Been Tested?

Tested all possible combos of the -Skip parameter including not using it and specifying all possible values combined (which will throw an error). Ensured that Files/SupportFiles is only copied and session property updated if those folders exist.

### DIFF
--- a/src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
+++ b/src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
@@ -20,6 +20,13 @@ function Copy-ADTContentToCache
     .PARAMETER LiteralPath
         The path to the software cache folder.
 
+    .PARAMETER Skip
+        Specifies one or more content categories to exclude from the cache copy. Acceptable values are 'Files', 'SupportFiles', and 'Other'.
+
+        - Files: Excludes the Files folder and does not remap the DirFiles session property.
+        - SupportFiles: Excludes the SupportFiles folder and does not remap the DirSupportFiles session property.
+        - Other: Excludes all content other than the Files and SupportFiles folders.
+
     .INPUTS
         None
 
@@ -34,6 +41,16 @@ function Copy-ADTContentToCache
         Copy-ADTContentToCache -LiteralPath "$envWinDir\Temp\PSAppDeployToolkit"
 
         This example copies the toolkit content to the specified cache folder.
+
+    .EXAMPLE
+        Copy-ADTContentToCache -Skip Files
+
+        This example copies the toolkit content to the default cache folder, excluding the Files folder and leaving DirFiles pointing at the original location.
+
+    .EXAMPLE
+        Copy-ADTContentToCache -Skip Other
+
+        This example copies only the Files and SupportFiles folders to the default cache folder, skipping all other content.
 
     .NOTES
         An active ADT session is required to use this function.
@@ -61,11 +78,20 @@ function Copy-ADTContentToCache
         [Parameter(Mandatory = $false)]
         [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
         [Alias('Path', 'PSPath')]
-        [System.String]$LiteralPath = "$((Get-ADTConfig).Toolkit.CachePath)\$((Get-ADTSession).InstallName)"
+        [System.String]$LiteralPath = "$((Get-ADTConfig).Toolkit.CachePath)\$((Get-ADTSession).InstallName)",
+
+        [Parameter(Mandatory = $false)]
+        [ValidateSet('Files', 'SupportFiles', 'Other')]
+        [System.String[]]$Skip
     )
 
     begin
     {
+        if ('Files' -in $Skip -and 'SupportFiles' -in $Skip -and 'Other' -in $Skip)
+        {
+            $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Skip -ProvidedValue $Skip -ExceptionMessage 'Cannot specify all possible values for -Skip parameter as there would be nothing to copy.'))
+        }
+
         try
         {
             $adtSession = Get-ADTSession
@@ -75,6 +101,7 @@ function Copy-ADTContentToCache
         {
             $PSCmdlet.ThrowTerminatingError($_)
         }
+
         Initialize-ADTFunction -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
     }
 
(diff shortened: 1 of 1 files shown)


src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -20,6 +20,13 @@ function Copy-ADTContentToCache
     .PARAMETER LiteralPath
         The path to the software cache folder.
 
+    .PARAMETER Skip
```
mjr4077au: In your [comment from September](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1740#issuecomment-3282617417), you opioned an `-Include` or _`-Exclude`_ parameter, for which I do think is the better name to use here.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -35,6 +42,16 @@ function Copy-ADTContentToCache
 
         This example copies the toolkit content to the specified cache folder.
 
+    .EXAMPLE
+        Copy-ADTContentToCache -Skip Files,SupportFiles
+
+        This example copies the toolkit content to the default cache folder, excluding the Files and SupportFiles folders and leaving DirFiles and DirSupportFiles pointing at the original location.
+
+    .EXAMPLE
+        Copy-ADTContentToCache -Skip Other
```
mjr4077au: I'm not sure how intuitive this will be for end users. In the aforementioned comment from September, you said:

> Also since the purpose here is to cache MSI content for future repairs, perhaps we should also stop copying the toolkit files and just cache the Files and/or SupportFiles folders only.

Given the name of the function is `Copy-ADTContentToCache`, I wholeheartedly support just the caching of data and all the toolkit files. Happy to discuss further if you've had a change of heart.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -61,11 +78,20 @@ function Copy-ADTContentToCache
...
+        [System.String]$LiteralPath = "$((Get-ADTConfig).Toolkit.CachePath)\$((Get-ADTSession).InstallName)",
+
+        [Parameter(Mandatory = $false)]
+        [ValidateSet('Files', 'SupportFiles', 'Other')]
+        [System.String[]]$Skip
     )
 
     begin
     {
+        if ('Files' -in $Skip -and 'SupportFiles' -in $Skip -and 'Other' -in $Skip)
```
mjr4077au: While not strictly necessary, some braces here wouldn't hurt.

```powershell
if (('Files' -in $Skip) -and ('SupportFiles' -in $Skip) -and ('Other' -in $Skip))
{
}
```

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -125,12 +152,41 @@ function Copy-ADTContentToCache
...
                     Copy-ADTFile -Path (Join-Path -Path $scriptDir -ChildPath *) -Destination $LiteralPath -Recurse
                 }
-                $adtSession.DirFiles = Join-Path -Path $LiteralPath -ChildPath Files
-                $adtSession.DirSupportFiles = Join-Path -Path $LiteralPath -ChildPath SupportFiles
+                else
+                {
+                    # Selective copy: enumerate top-level items and copy based on -Skip.
+                    if ('Other' -notin $Skip)
+                    {
+                        Get-ChildItem -LiteralPath $scriptDir -Force | & { process { if ($_.Name -notin @('Files', 'SupportFiles')) { Copy-ADTFile -LiteralPath $_.FullName -Destination $LiteralPath -Recurse } } }
```
mjr4077au: Putting `@('Files', 'SupportFiles')` in the process block is creating this array every time, and there could be thousands of files coming through the pipe. Add a `begin {}` block and place it there.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -125,12 +152,41 @@ function Copy-ADTContentToCache
...
+                    }
+                    $filesSourcePath = Join-Path -Path $scriptDir -ChildPath Files
+                    if ('Files' -notin $Skip -and (Test-Path -LiteralPath $filesSourcePath -PathType Container))
```
mjr4077au: Braces here would be nice also, but just a nit.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -125,12 +152,41 @@ function Copy-ADTContentToCache
...
+                    {
+                        Copy-ADTFile -LiteralPath $filesSourcePath -Destination $LiteralPath -Recurse
+                    }
+                    $supportFilesSourcePath = Join-Path -Path $scriptDir -ChildPath SupportFiles
+                    if ('SupportFiles' -notin $Skip -and (Test-Path -LiteralPath $supportFilesSourcePath -PathType Container))
```
mjr4077au: Braces here would be nice also, but just a nit.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -125,12 +152,41 @@ function Copy-ADTContentToCache
...
+                    {
+                        Copy-ADTFile -LiteralPath $supportFilesSourcePath -Destination $LiteralPath -Recurse
+                    }
+                }
+
+                # Remap session properties for categories that were copied.
+                $filesDestPath = Join-Path -Path $LiteralPath -ChildPath Files
+                if ('Files' -notin $Skip -and (Test-Path -LiteralPath $filesDestPath -PathType Container))
```
mjr4077au: Braces here would be nice also, but just a nit.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -125,12 +152,41 @@ function Copy-ADTContentToCache
...
+                {
+                    $adtSession.DirFiles = $filesDestPath
+                }
+                $supportFilesDestPath = Join-Path -Path $LiteralPath -ChildPath SupportFiles
+                if ('SupportFiles' -notin $Skip -and (Test-Path -LiteralPath $supportFilesDestPath -PathType Container))
```
mjr4077au: Braces here would be nice also, but just a nit.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
mjr4077au: On second thought, maybe just check that the length isn't 3?

mjr4077au (changes requested):
Please have a look at my review comments and let me know if you've got any questions. I suppose as a last matter, does anything need to be done in `Remove-ADTContentFromCache` to match up?

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: I opted not to go this way as Exclude params typically accept wildcard filters (such as in Get-Item) which would result in a more complex command for the user (e.g `-Exclude "$($adtSession.DirFiles)\*"` vs `-Skip 'Files'`).

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: It feels pretty simple and intuitive to me, and it matches the user's request. There are valid use cases for excluding each type of content:

- Skip Files/SupportFiles if you want to point your uninstall string to a cached copy of Invoke-AppDeployToolkit.exe, and you don't need the installer files there for uninstall

- Skip Other if you just want to keep a copy of the Files content for MSI repair purposes

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: They weren't added as they're not needed, but if you find setups like that easier to read, we can put them in.

I thought about the count approach but that wouldn't work if someone did -Skip Files,Files,Files

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: One would hope there wouldnt't be thousands in the top level (there's no recurse on there), but good point, and added.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: I don't think we need to do anything to `Remove-ADTContentFromCache`, it can just blat the entire folder contents as it always has.

mjr4077au: You don't have to use wildcards with `-Exclude` on Get-Item, it's just that you can if you wish to. I do think it's the better parameter name to use here.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
mjr4077au: Maybe "Toolkit" is a better name to use here?

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: As mentioned in https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1740, I've switched to -Exclude, but opted not to go the wildcard route as we want to avoid partial copying of Files since that could break things.

mjr4077au: Yeah agreed, I never wanted wildcards here anyway.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
DanGough: Ok, done

mjr4077au (approved):

Issue #2074, opened 2026-03-22
[Bug] Get-ADTWindowTitle -WindowTitle regex matching doesn't work


obuolinis:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.1.8

### Describe the bug

I was converting a package from v3 and noticed `Get-ADTWindowTitle -WindowTitle` doesn't seem to accept regex the way v3 cmdlet did. I was simply searching for `window.+title` and v4 cmdlet didn't produce any matches. It matched separate words 'window' and 'title' though. Then i tested a few more simple patterns and v4 didn't match any. So I guess it processes the value as a literal string, not a regex as declared.
Same obviously applies to Send-ADTKeys.

### Steps to reproduce

1.
2.
3.

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 24H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 26100.1.amd64fre.ge_release.240331-1435
OsLanguage           : en-GB
OsMuiLanguages       : {en-GB, en-US}
KeyboardLayout       : en-US
TimeZone             : (UTC+02:00) Helsinki, Kyiv, Riga, Sofia, Tallinn, Vilnius
HyperVisorPresent    : True
CsPartOfDomain       : False
CsPCSystemType       : Mobile
```

obuolinis:
Also have to add that usage examples on the website for the function Get-ADTWindowTitle are outdated - parameter `-GetAllWindowTitles` is no longer supported as far as i can tell.

mjr4077au:
The issues with the `-WindowTitle` have been addressed for 4.2.0 already, can you try a current development build to confirm? https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/actions/runs/23408577310

Regarding `-GetAllWindowTitles`, just don't specify anything and all windows will be returned.

obuolinis:
[...]

PR #2078, opened 2026-03-24
Another smattering of minor touch ups, mostly from CodeQL reports that annoyingly only present themselves when PRs are merged into main...

mjr4077au:
# Pull Request

## Description

As per the title, just a number of annoying nits that CodeQL mentions when new code gets merged into main. All are valid though and make sense.

Other non-CodeQL improvements are:

* Adjust the logic in `Get-ADTSessionCacheScriptDirectory` to match `Open-ADTSession`.
* Don't allow `Resolve-ADTErrorRecord` to print the PositionMessage in `Start-ADTProcess` if `-SecureArgumentList` was specified.

Issues closed by this PR:

* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/2074.

## Type of change

## Checklist

### DIFF
--- a/src/PSAppDeployToolkit/Public/Start-ADTMsiProcess.ps1
+++ b/src/PSAppDeployToolkit/Public/Start-ADTMsiProcess.ps1
@@ -213,13 +213,11 @@ function Start-ADTMsiProcess
 
         [Parameter(Mandatory = $false)]
         [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
-        [PSAppDeployToolkit.Attributes.ValidateUnique()]
         [PSDefaultValue(Help = 'Install (Normal): (Get-ADTConfig).MSI.InstallParams; Install (Silent): (Get-ADTConfig).MSI.SilentParams; Uninstall (Normal): (Get-ADTConfig).MSI.UninstallParams; Uninstall (Silent): (Get-ADTConfig).MSI.SilentParams')]
         [System.String[]]$ArgumentList,
 
         [Parameter(Mandatory = $false)]
         [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
-        [PSAppDeployToolkit.Attributes.ValidateUnique()]
         [System.String[]]$AdditionalArgumentList,
 
         [Parameter(Mandatory = $false)]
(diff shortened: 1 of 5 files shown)


src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -87,9 +87,9 @@ function Copy-ADTContentToCache
 
     begin
     {
-        if ('Files' -in $Exclude -and 'SupportFiles' -in $Exclude -and 'Toolkit' -in $Exclude)
+        if (($null -ne $Exclude) -and ($Exclude.Count -eq $MyInvocation.MyCommand.Parameters.Exclude.Attributes.Where({$_ -is [System.Management.Automation.ValidateSetAttribute]}).ValidValues.Count))
```
DanGough: As mentioned before, a basic count setup is not good enough here as it will reject an input like `-Exclude Files,Files,Files`.

I'd do something like:

`($Exclude | Select-Object -Unique -CaseInsensitive).Count`

mjr4077au: @DanGough fair callout, I missed that in the last PR. I've added a commit to address that. There's no `-CaseInsensitive` parameter in Windows PowerShell, but it shouldn't be needed here either.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -88,7 +88,7 @@ function Copy-ADTContentToCache
     begin
     {
         Initialize-ADTFunction -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
-        if (($null -ne $Exclude) -and ($Exclude.Count -eq $MyInvocation.MyCommand.Parameters.Exclude.Attributes.Where({ $_ -is [System.Management.Automation.ValidateSetAttribute] }).ValidValues.Count))
+        if (($null -ne $Exclude) -and (($Exclude | Select-Object -Unique | Measure-Object).Count -eq $MyInvocation.MyCommand.Parameters.Exclude.Attributes.Where({ $_ -is [System.Management.Automation.ValidateSetAttribute] }).ValidValues.Count))
```
DanGough: I know this is unlikely to ever be an issuein reality, but TIL that:

PS5 has a -CaseSensitive option and is case insensitive by default
PS7 has a -CaseInsensitive option and is case sensitive by default 

🤦‍♂️

mjr4077au: This has been addressed with successive commits to implement new functionality for uniqueness.

PR #2090, opened 2026-03-26
New capabilities for New-ADTTemplate

### LINKED ISSUE #1099: [Feature] v4 New-ADTTemplate populates the $adtsession values

As a wrapper/ packager, I want to execute New-ADTTemplate with parameters that populates the $adtsession values, so that it can be used on an automated scaffold process or CI/CD pipeline.

```Powershell
New-ADTTemplate -Destination c:\temp -Name 'dbeaver-ce-24.3.0-win32.win32.x86_64' -Version 4 -AppVendor dbeaver  -AppName  'DBeaver - Community Edition' -AppVersion 24.3.0 -AppArch x64 -AppLang EN -AppRevision 01 -AppSuccessExitCodes '0' -AppRebootExitCodes '1641,3010' -AppScriptAuthor 'Humble Wrapper'
```

or the variables be populated by a psd1, 

```Powershell
New-ADTTemplate -Destination c:\temp -Name 'dbeaver-ce-24.3.0-win32.win32.x86_64' -Version 4 -ADTSessionPs1Path c:\temp\dbeaver-ce-24.3.0-win32.win32.x86_64.psd1
```

DanGough:
# Pull Request

## Description

New parameters for New-ADTTemplate:
```
    .PARAMETER SessionProperties
        A dictionary of key-value pairs to inject into the $adtSession hashtable of the generated Invoke-AppDeployToolkit.ps1. Accepts [hashtable], [ordered], or any [System.Collections.IDictionary] type. Only supported when -Version is 4.

    .PARAMETER Config
        A dictionary of key-value pairs to override in the generated Config\config.psd1. The dictionary structure must mirror the config file's nested hashtable layout (e.g. @{ MSI = @{ InstallParams = 'REBOOT=ReallySuppress /QB-!' } }). Only existing keys can be overridden; specifying a key that does not exist in the default config will throw an error.

    .PARAMETER PreInstallScriptBlock
        A ScriptBlock whose content will replace the Pre-Install phase of the Install-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER InstallScriptBlock
        A ScriptBlock whose content will replace the Install phase of the Install-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER PostInstallScriptBlock
        A ScriptBlock whose content will replace the Post-Install phase of the Install-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER PreUninstallScriptBlock
        A ScriptBlock whose content will replace the Pre-Uninstall phase of the Uninstall-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER UninstallScriptBlock
        A ScriptBlock whose content will replace the Uninstall phase of the Uninstall-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER PostUninstallScriptBlock
        A ScriptBlock whose content will replace the Post-Uninstall phase of the Uninstall-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER PreRepairScriptBlock
        A ScriptBlock whose content will replace the Pre-Repair phase of the Repair-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER RepairScriptBlock
        A ScriptBlock whose content will replace the Repair phase of the Repair-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER PostRepairScriptBlock
        A ScriptBlock whose content will replace the Post-Repair phase of the Repair-ADTDeployment function in the generated Invoke-AppDeployToolkit.ps1. Only supported when -Version is 4.

    .PARAMETER Assets
        An array of file or folder paths to copy into the Assets folder of the generated template. Paths are passed to Copy-Item with -Recurse.

    .PARAMETER Files
        An array of file or folder paths to copy into the Files folder of the generated template. Paths are passed to Copy-Item with -Recurse.

    .PARAMETER SupportFiles
        An array of file or folder paths to copy into the SupportFiles folder of the generated template. Paths are passed to Copy-Item with -Recurse.

    .EXAMPLE
        New-ADTTemplate -Destination 'C:\Temp' -Assets 'C:\Assets\AppIconLight.png', 'C:\Assets\AppIconDark.png' -Config @{ Assets = @{ Logo = '..\Assets\AppIconLight.png'; LogoDark = '..\Assets\AppIconDark.png' } }

    .EXAMPLE
        New-ADTTemplate -Destination 'C:\Temp' -SessionProperties @{ AppVendor = 'Contoso'; AppName = 'MyApp'; AppVersion = '6.7' } -Files 'C:\Installers\Setup.msi' -InstallScriptBlock { Start-ADTMsiProcess -Action Install -FilePath 'Setup.msi' } -UninstallScriptBlock {  Start-ADTMsiProcess -Action Uninstall -FilePath 'Setup.msi' }
```
This effectively lets you generate an entire deployment package by specifying the assets, config, files. and script snippets:

```powershell
$params = @{
    SessionProperties = @{
        AppVendor = 'Contoso'
        AppName = 'MyApp'
        AppVersion = '6.7'
        RequireAdmin = $false
        AppProcessesToClose = @(
            'notepad'
            @{
                Name = 'calc'
                Description = 'Calculator'
            }
        )
    }
    Config = @{
        Assets = @{
            Logo = '..\Assets\AppIconLight.png'
            LogoDark = '..\Assets\AppIconDark.png'
        }
        Toolkit = @{
            LogPath = '$env:ProgramData\Microsoft\IntuneManagementExtension\Logs'
        }
    }
    Assets = 'C:\Assets\AppIconLight.png', 'C:\Assets\AppIconDark.png'
    Files = 'C:\Installers\Setup.msi'
    InstallScriptBlock = {
        Start-ADTMsiProcess -Action Install -FilePath 'Setup.msi'
    }
    UninstallScriptBlock = {
        Start-ADTMsiProcess -Action Uninstall -FilePath 'Setup.msi'
    }
}

New-ADTTemplate @params
```

Resolves: #1099

## Type of change

## Checklist

## How Has This Been Tested?

Manual testing + Pester tests.


DanGough:
Proposed further changes after dicussion with @mjr4077au:

- Remove restrictions on config settings that prevent users specifying options that do not exist in config.psd1; users are free to extend config and some have already done this in their extensions
- Make the quote handling code in the expression generator function more robust, and add support for more types such as scriptblocks, since users can add their own custom properties here. Investigate if we can access the raw input into the function instead of only looking at the bound parameters and trying to convert that to PowerShell script.
- Modify the template to use separate scriptblocks for each section so that we don't need to rely on specific comments/markers for text replacements. Scriptblocks make more sense than functions here to avoid any scoping issues with local variables created inside functions.
- Consider adding an auto-repair feature that will be able to run uninstall-then-install if repair code is undefined. If the install phase is set outside of these scriptblocks then the phase will still be correctly set as Repair throughout.
- Consider removing ZeroConfig code from default template and adding a ZeroConfig option to the function that will inject it. This could then be added to the build system so that a separate ZeroConfig template is generated as an artifact.

DanGough:
> * Consider adding an auto-repair feature that will be able to run uninstall-then-install if repair code is undefined. If the install phase is set outside of these scriptblocks then the phase will still be correctly set as Repair throughout.

All implemented apart from the above, which was a nice-to-have anyway. This one is a bit finicky as you might want to run:

Pre-Uninstall -> Uninstall -> Post-Uninstall -> Pre-Install -> Install -> Post-Install

This sequence might result in duplicate welcome prompts and/or "<deploymentphase> complete!" prompts, so would need some further thought. We can make this a feature further down the line if needed.

DanGough:
Build is failing after rebasing due to issue on main - rebasing this again after https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/pull/2108 is merged should resolve this.

src/PSAppDeployToolkit.Build/Private/Export-ADTScriptTemplate.ps1
```diff
@@ -17,9 +17,14 @@ function Export-ADTScriptTemplate
     try
     {
         Write-ADTBuildLogEntry -Message "Creating frontend templates, this may take a while."
+        $templateArgs = @(
+            "-Name 'Template_v3' -Version 3 -Force"
```
mjr4077au: Nit, but the `-Force` could be moved out of here and back into the main line below.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -1,62 +1,45 @@
...
-The script imports the PSAppDeployToolkit module which contains the logic and functions required to install or uninstall an application.
+The script imports the PSAppDeployToolkit module and executes the specified install/uninstall/repair scriptblocks.
 
 .PARAMETER DeploymentType
-The type of deployment to perform.
+The type of deployment to perform, Install, Uninstall, or Repair. Default is Install.
 
 .PARAMETER DeployMode
 Specifies whether the installation should be run in Interactive (shows dialogs), Silent (no dialogs), NonInteractive (dialogs without prompts) mode, or Auto (shows dialogs if a user is logged on, device is not in the OOBE, and there's no running apps to close).
-
```
mjr4077au, line 18: Would probably prefer this line feed be added back in as it makes it nicer to read in an editor.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -1,62 +1,45 @@
...
-- 69000 - 69999: Recommended for user customized exit codes in Invoke-AppDeployToolkit.ps1
-- 70000 - 79999: Recommended for user customized exit codes in PSAppDeployToolkit.Extensions module.
-
 .LINK
 https://psappdeploytoolkit.com
 
 #>
 
 [CmdletBinding()]
+[System.Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', 'PreInstall', Justification = 'This variable is accessed dynamically via Get-Variable and therefore cannot be seen by PSScriptAnalyzer.')]
```
mjr4077au, line 30: These suppression attributes get emitted into the template as per `New-ADTTemplate`, which should be addressed.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -96,40 +72,23 @@ $adtSession = @{
...
     AppSuccessExitCodes = @(0)
     AppRebootExitCodes = @(1641, 3010)
-    AppProcessesToClose = @()  # Example: @('excel', @{ Name = 'winword'; Description = 'Microsoft Word' })
+    AppProcessesToClose = @()
     AppScriptVersion = '1.0.0'
     AppScriptDate = '2000-12-31'
     AppScriptAuthor = '<author name>'
     RequireAdmin = $true
 
-    # Install Titles (Only set here to override defaults set by the toolkit).
```
mjr4077au, line 104: I think this comment should be restored, or the the `InstallName`/`InstallTitle` lines be removed entirely. Without the comment, people will think they should be setting them, which they should not be.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
     }
     Show-ADTInstallationWelcome @saiwParams
```
mjr4077au: Could probably remove this blank line if you're desperate to keep the line count of the template down.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
...
-        if ($adtSession.DefaultMspFiles)
-        {
-            $adtSession.DefaultMspFiles | Start-ADTMsiProcess -Action Patch
-        }
-    }
-
-    ## <Perform Installation tasks here>
 
+## MARK: Install
+$Install = {
```
mjr4077au: These blank line feeds in the empty scriptblocks should be removed as they harm the diff output for callers who might be using Git to track their changes.

For instance, you can see below that editing the line to add some code constitutes a line deletion, then one or more additions:
Image

Below is when the blank line isn't there:
Image

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
...
 
-    ##================================================
-    ## MARK: Post-Install
-    ##================================================
-    $adtSession.InstallPhase = "Post-$($adtSession.DeploymentType)"
+}
 
-    ## <Perform Post-Installation tasks here>
```
mjr4077au: Could probably remove this blank line if you're desperate to keep the line count of the template down.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -96,40 +72,23 @@ $adtSession = @{
     AppRevision = '01'
     AppSuccessExitCodes = @(0)
     AppRebootExitCodes = @(1641, 3010)
-    AppProcessesToClose = @()  # Example: @('excel', @{ Name = 'winword'; Description = 'Microsoft Word' })
+    AppProcessesToClose = @()
     AppScriptVersion = '1.0.0'
     AppScriptDate = '2000-12-31'
     AppScriptAuthor = '<author name>'
     RequireAdmin = $true
```
mjr4077au: Thoughts around getting rid of this `RequireAdmin = $true` line entirely? I personally don't like it and find it annoying more than anything else.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -96,40 +72,23 @@ $adtSession = @{
     AppRevision = '01'
     AppSuccessExitCodes = @(0)
     AppRebootExitCodes = @(1641, 3010)
-    AppProcessesToClose = @()  # Example: @('excel', @{ Name = 'winword'; Description = 'Microsoft Word' })
```
mjr4077au, line 99: Given our user base, I'm not sure we can get away without some kind of example comment here. Even _with_ the comment, people still cock this up.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -1,62 +1,45 @@
...
+The type of deployment to perform, Install, Uninstall, or Repair. Default is Install.
 
 .PARAMETER DeployMode
 Specifies whether the installation should be run in Interactive (shows dialogs), Silent (no dialogs), NonInteractive (dialogs without prompts) mode, or Auto (shows dialogs if a user is logged on, device is not in the OOBE, and there's no running apps to close).
-
 Silent mode is automatically set if it is detected that the process is not user interactive, no users are logged on, the device is in Autopilot mode, or there's specified processes to close that are currently running.
 
 .PARAMETER SuppressRebootPassThru
-Suppresses the 3010 return code (requires restart) from being passed back to the parent process (e.g. SCCM) if detected from an installation. If 3010 is passed back to SCCM, a reboot prompt will be triggered.
+Suppresses the 3010 return code (requires restart) from being passed back to the parent process (e.g. Intune) if detected from an installation.
```
mjr4077au: This should probably be changed to "Prevents the toolkit from exiting with a defined reboot exit code", or something along these lines. While it defaults to 3010, its completely customisable.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -1,62 +1,45 @@
...
 
 .PARAMETER TerminalServerMode
```
mjr4077au, line 19: Thoughts around dropping this parameter entirely? Do we need it? Is anyone using it?

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -1,62 +1,45 @@
...
 Changes to "user install mode" and back to "user execute mode" for installing/uninstalling applications for Remote Desktop Session Hosts/Citrix servers.
 
 .PARAMETER DisableLogging
```
mjr4077au, line 22: Thoughts around dropping this parameter entirely? Do we need it? Is anyone using it?

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -1,62 +1,45 @@
...
-.OUTPUTS
-None. This script does not generate any output.
-
-.NOTES
-Toolkit Exit Code Ranges:
-- 60000 - 68999: Reserved for built-in exit codes in Invoke-AppDeployToolkit.ps1, and Invoke-AppDeployToolkit.exe
-- 69000 - 69999: Recommended for user customized exit codes in Invoke-AppDeployToolkit.ps1
-- 70000 - 79999: Recommended for user customized exit codes in PSAppDeployToolkit.Extensions module.
-
 .LINK
```
mjr4077au: Can probably drop the Link section from the comment-based help as well.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
     }
     Show-ADTInstallationWelcome @saiwParams
 
-    ## Show Progress Message (with the default message).
     Show-ADTInstallationProgress
+}
```
mjr4077au, line 102: We could probably have a single blank line between Deployment sections, then leave the double space for Install/Uninstall/Repair breaks.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
...
 
 
-##================================================
 ## MARK: Initialization
-##================================================
-
-# Set strict error handling across entire operation.
 $ErrorActionPreference = [System.Management.Automation.ActionPreference]::Stop
 $ProgressPreference = [System.Management.Automation.ActionPreference]::SilentlyContinue
 Set-StrictMode -Version 1
```
mjr4077au: Can delete this space and just run all the code together.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
...
     {
-        Get-ChildItem -LiteralPath "$PSScriptRoot\..\..\..\..\PSAppDeployToolkit" -Recurse -File | Unblock-File -ErrorAction Ignore
-        Import-Module -FullyQualifiedName @{ ModuleName = [System.Management.Automation.WildcardPattern]::Escape("$PSScriptRoot\..\..\..\..\PSAppDeployToolkit\PSAppDeployToolkit.psd1"); Guid = '8c3c366b-8606-4576-9f2d-4051144f7ca2'; ModuleVersion = '4.2.0' } -Force
+        Get-ChildItem -LiteralPath $moduleBase -Recurse -File | Unblock-File -ErrorAction Ignore
+        Import-Module -FullyQualifiedName @{ ModuleName = [System.Management.Automation.WildcardPattern]::Escape("$moduleBase\PSAppDeployToolkit.psd1"); Guid = '8c3c366b-8606-4576-9f2d-4051144f7ca2'; ModuleVersion = '4.2.0' } -Force
     }
     else
     {
         Import-Module -FullyQualifiedName @{ ModuleName = 'PSAppDeployToolkit'; Guid = '8c3c366b-8606-4576-9f2d-4051144f7ca2'; ModuleVersion = '4.2.0' } -Force
     }
```
mjr4077au: Can delete this space and just run all the code together.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -142,176 +101,86 @@ function Install-ADTDeployment
...
 $ErrorActionPreference = [System.Management.Automation.ActionPreference]::Stop
 $ProgressPreference = [System.Management.Automation.ActionPreference]::SilentlyContinue
 Set-StrictMode -Version 1
 
-# Import the module and instantiate a new session.
 try
 {
-    # Import the module locally if available, otherwise try to find it from PSModulePath.
-    if (Test-Path -LiteralPath "$PSScriptRoot\..\..\..\..\PSAppDeployToolkit\PSAppDeployToolkit.psd1" -PathType Leaf)
+    $moduleBase = "$PSScriptRoot\..\..\..\..\PSAppDeployToolkit"
```
mjr4077au: The only reason this was never a variable before was that I was trying to not have the script create variables in the script's scope that are irrelevant to the caller. The only variables the script was making before was `$adtSession` and `$iadtParams`, and with the latter I did consider adding a `Remove-Variable` line to get rid of it.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -323,14 +192,10 @@ catch
...
 
-##================================================
 ## MARK: Invocation
-##================================================
-
-# Commence the actual deployment operation.
 try
 {
-    # Import any found extensions before proceeding with the deployment.
+    # Import any PSAppDeployToolkit.* extensions
```
mjr4077au, line 175: If we're getting rid of all the comments, why was this one kept?

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -342,21 +207,29 @@ try
         }
     }
```
mjr4077au: Can delete this space and just run all the code together.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -342,21 +207,29 @@ try
...
 
-    # Invoke the deployment and close out the session.
-    & "$($adtSession.DeploymentType)-ADTDeployment"
+    foreach ($prefix in 'Pre-', '', 'Post-')
+    {
+        $installPhase = "$prefix$($adtSession.DeploymentType)"
+        $scriptBlock = Get-Variable -Name $installPhase.Replace('-', '') -ValueOnly -ErrorAction Ignore
```
mjr4077au: Nit, but `.Replace('-', $null)` would probably be preferable here.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -342,21 +207,29 @@ try
...
+        if (![System.String]::IsNullOrWhiteSpace($scriptBlock))
+        {
+            $adtSession.InstallPhase = $installPhase
+            . $scriptBlock
+        }
+    }
+
```
mjr4077au: Can delete this space and just run all the code together.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -342,21 +207,29 @@ try
...
 catch
 {
-    # An unhandled error has been caught.
     $mainErrorMessage = "An unhandled error within [$($MyInvocation.MyCommand.Name)] has occurred.`n$(Resolve-ADTErrorRecord -ErrorRecord $_)"
     Write-ADTLogEntry -Message $mainErrorMessage -Severity Error
 
     ## Error details hidden from the user by default. Show a simple dialog with full stack trace:
     # Show-ADTDialogBox -Text $mainErrorMessage -Icon Stop -NoWait
 
     ## Or, a themed dialog with basic error message:
```
mjr4077au, line 206: Thoughts around getting of all these commented out potential examples in this catch block? I think overall, they make the block look messier with their presence.
Image

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
```diff
@@ -342,21 +207,29 @@ try
...
-    # Show-ADTInstallationPrompt -Message "$($adtSession.DeploymentType) failed at line $($_.InvocationInfo.ScriptLineNumber), char $($_.InvocationInfo.OffsetInLine):`n$($_.InvocationInfo.Line.Trim())`n`nMessage:`n$($_.Exception.Message)" -ButtonRightText OK -Icon Error -NoWait
+    # Show-ADTInstallationPrompt -Message "$($adtSession.DeploymentType) failed at line $($_.InvocationInfo.ScriptLineNumber), char $($_.InvocationInfo.OffsetInLine):`n$($_.InvocationInfo.Line.Trim())`n`nMessage:`n$($_.Exception.Message)" -ButtonRightText OK -NoWait
 
     Close-ADTSession -ExitCode 60001
 }
```
mjr4077au, line 210: Just a comment for reference, but if the line feed additions/removals, etc I'm proposing get implemented, the out of box template would be 195 lines:
Image

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
         [System.Int32]$Version = 4,
 
+        [Parameter(Mandatory = $false)]
+        [System.Collections.IDictionary]$SessionProperties,
+
+        [Parameter(Mandatory = $false)]
+        [System.Collections.IDictionary]$Config,
+
+        [Parameter(Mandatory = $false)]
+        [System.Management.Automation.ScriptBlock]$PreInstallScriptBlock,
```
mjr4077au, line 163: Please add `[PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]` to all these scriptblock parameters.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
         [ValidateRange(3, 4)]
         [System.Int32]$Version = 4,
 
+        [Parameter(Mandatory = $false)]
+        [System.Collections.IDictionary]$SessionProperties,
```
mjr4077au, line 155: Please add `[PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]` to all these dictionary parameters.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+
+        [Parameter(Mandatory = $false)]
+        [System.Management.Automation.ScriptBlock]$RepairScriptBlock,
+
+        [Parameter(Mandatory = $false)]
+        [System.Management.Automation.ScriptBlock]$PostRepairScriptBlock,
+
+        [Parameter(Mandatory = $false)]
+        [SupportsWildcards()]
+        [System.String[]]$Assets,
```
mjr4077au, line 200: Please add `[PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]` to all these `string[]` parameters.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
 
         [Parameter(Mandatory = $false)]
         [System.Management.Automation.SwitchParameter]$Force,
 
         [Parameter(Mandatory = $false)]
-        [System.Management.Automation.SwitchParameter]$PassThru
+        [System.Management.Automation.SwitchParameter]$PassThru,
+
+        [Parameter(Mandatory = $false)]
+        [System.Management.Automation.SwitchParameter]$ZeroConfig
```
mjr4077au: I think it's better that `-ZeroConfig` comes before `-PassThru`. Typically, `-PassThru` is the last parameter you define.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+                if ($InputObject) { return '$true' } else { return '$false' }
+            }
+            if ($InputObject -is [System.DateTime] -or $InputObject -is [System.DateTimeOffset])
+            {
+                return "(Get-Date '$($InputObject.ToString('o'))')"
+            }
+            if ($InputObject -is [System.TimeSpan])
+            {
+                # Look at Days/Hours/Minutes/Seconds/Milliseconds properties in ascending order. whichever one is the first non-zero is the most concise format to represent the TimeSpan. Output a string of the form (New-Timespan -Minutes 5) or (New-Timespan -Seconds 30) etc. Include the parentheses in the string.
+                $timeSpanArgs = foreach ($property in 'Days', 'Hours', 'Minutes', 'Seconds', 'Milliseconds')
```
mjr4077au: This is really over-complicated. Why not just do `return "[System.TimeSpan]'$($InputObject.ToString())'"` at the end?

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+            }
+
+            $naerParams = @{
+                Exception = [System.ArgumentException]::new("Session property value of type [$($InputObject.GetType().FullName)] is not supported.")
+                Category = [System.Management.Automation.ErrorCategory]::InvalidArgument
+                ErrorId = 'UnsupportedSessionPropertyValueType'
+                TargetObject = $InputObject
+                RecommendedAction = 'Use any System.ValueType, strings, datetimes, timespans, scriptblocks, hashtables, or ordered dictionaries.'
+            }
+            throw (New-ADTErrorRecord @naerParams)
```
mjr4077au: This should use `$PSCmdlet.ThrowTerminatingError()` instead of `throw`.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+            [OutputType([System.String])]
+            param
+            (
+                [Parameter(Mandatory = $true)]
+                [System.Management.Automation.ScriptBlock[]]$ScriptBlock
+            )
+
+            $scriptBodies = foreach ($sb in $ScriptBlock)
+            {
+                $lines = [regex]::Replace(($sb.ToString() -replace '\r?\n', [System.Environment]::NewLine).TrimStart([System.Environment]::NewLine).TrimEnd(), '(?m)^\t+', { param($m) '    ' * $m.Value.Length }) -split [System.Environment]::NewLine
```
mjr4077au: There's a lot of regex throughout this function which is slow and can be hard to read as well, particularly as there's a mix of PowerShell regex operators like `-replace` and dotnet operators like the `[regex]` class.

If you look at [`Export-ADTScriptBlockToFile`](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/main/src/PSAppDeployToolkit/Private/Export-ADTScriptBlockToFile.ps1), you'll see we're achieving this without any regex at all, and not only that, a singular loop.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+
+        # Helper to normalize (and optionally concatenate) scriptblocks. Each input is trimmed, line endings normalized, source indentation stripped, leading tabs replaced with spaces.
+        function ConvertTo-ADTScriptBody
+        {
+            [CmdletBinding()]
+            [OutputType([System.String])]
+            param
+            (
+                [Parameter(Mandatory = $true)]
+                [System.Management.Automation.ScriptBlock[]]$ScriptBlock
```
mjr4077au, line 325: This should have a `[PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]` on it as well.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+                return '$null'
+            }
+            if (($InputObject -is [System.String]) -or ($InputObject -is [System.Char]) -or ($InputObject -is [System.Version]) -or ($InputObject -is [System.Guid]) -or ($InputObject -is [System.IO.FileSystemInfo]) -or ($InputObject.GetType().IsEnum))
+            {
+                $str = $InputObject.ToString()
+                if (!$LiteralString -and ($str -match '(?<!`)\$'))
+                {
+                    return "`"$($str -replace '(?<!`)"', '`"')`""
+                }
+                return "'$($str.Replace("'", "''"))'"
```
mjr4077au: Best practice would be to use `[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent()` here.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+
+                [Parameter(Mandatory = $false)]
+                [System.Management.Automation.SwitchParameter]$LiteralString
+            )
+
+            if ($null -eq $InputObject)
+            {
+                return '$null'
+            }
+            if (($InputObject -is [System.String]) -or ($InputObject -is [System.Char]) -or ($InputObject -is [System.Version]) -or ($InputObject -is [System.Guid]) -or ($InputObject -is [System.IO.FileSystemInfo]) -or ($InputObject.GetType().IsEnum))
```
mjr4077au: This entire setup would be better as a switch since all if statement conditionals are based on type.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -79,21 +144,220 @@ function New-ADTTemplate
...
+                        $ExecuteDefaultMSISplat.Add('Transforms', $adtSession.DefaultMstFile)
+                    }
+                    Start-ADTMsiProcess @ExecuteDefaultMSISplat
+                }
+            }
+            foreach ($sbName in 'InstallScriptBlock', 'UninstallScriptBlock', 'RepairScriptBlock')
+            {
+                if ($PSBoundParameters.ContainsKey($sbName))
+                {
+                    $PSBoundParameters[$sbName] = [System.Management.Automation.ScriptBlock]::Create((ConvertTo-ADTScriptBody -ScriptBlock $zeroConfigScriptBlock, $PSBoundParameters[$sbName]))
```
mjr4077au, line 488: The `$PSBoundParameters.Add()` method is preferred over index additions as it'll throw if you're trying to add a value with a pre-existing key.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -131,7 +395,7 @@ function New-ADTTemplate
                         TargetObject = $badFiles
                         RecommendedAction = "Please re-download $($MyInvocation.MyCommand.Module.Name) and try again."
                     }
-                    $PSCmdlet.ThrowTerminatingError((New-ADTErrorRecord @naerParams))
+                    throw (New-ADTErrorRecord @naerParams)
```
mjr4077au: Why?

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -158,10 +422,97 @@ function New-ADTTemplate
...
+                if ($PSBoundParameters.ContainsKey('Config') -and $Config.Count -gt 0)
+                {
+                    $configText = $configBlock.ToString()
+                    $configAst = [System.Management.Automation.Language.Parser]::ParseInput($configText, [ref]$null, [ref]$null)
+                    $configHashtableAst = $configAst.EndBlock.Statements[0].PipelineElements[0].Expression
+                    $configReplacements = [System.Collections.Generic.List[PSCustomObject]]::new()
+                    $configInsertions = @{}
+
+                    # Recursive scriptblock to walk user dictionary against config AST.
+                    $collectReplacements = {
```
mjr4077au: This should be a function in the `begin` block like the others you added. Its definition mid-way through what's now a long function impacts readability.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -158,10 +422,97 @@ function New-ADTTemplate
...
+                [System.IO.File]::WriteAllBytes("$templatePath\Assets\Banner.Classic.png", [System.Convert]::FromBase64String(($banner = $defaultAssets.Where({ $_.Item1.Value.Equals('Banner') }).Item2.PipelineElements.Expression.Value)))
+                [System.IO.File]::WriteAllBytes("$templatePath\Assets\AppIcon.ico", [System.Convert]::FromBase64String(($logo = $defaultAssets.Where({ $_.Item1.Value.Equals('Logo') }).Item2.PipelineElements.Expression.Value)))
+                $configBlock = [System.Management.Automation.ScriptBlock]::Create($ADT.ModuleDefaults.Config.([System.String]::Empty).ToString().Replace($banner, '..\Assets\Banner.Classic.png').Replace($logo, '..\Assets\AppIcon.ico'))
+
+                # Override config values if specified.
+                if ($PSBoundParameters.ContainsKey('Config') -and $Config.Count -gt 0)
+                {
+                    $configText = $configBlock.ToString()
+                    $configAst = [System.Management.Automation.Language.Parser]::ParseInput($configText, [ref]$null, [ref]$null)
```
mjr4077au: Why not just `$configBlock.Ast` instead of all this?

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -158,10 +422,97 @@ function New-ADTTemplate
...
+                $defaultAssets = $Script:ADT.ModuleDefaults.Config.([System.String]::Empty).Ast.EndBlock.Statements.PipelineElements.Expression.KeyValuePairs.Where({ $_.Item1.Value.Equals('Assets') }).Item2.PipelineElements.Expression.KeyValuePairs
+                [System.IO.File]::WriteAllBytes("$templatePath\Assets\Banner.Classic.png", [System.Convert]::FromBase64String(($banner = $defaultAssets.Where({ $_.Item1.Value.Equals('Banner') }).Item2.PipelineElements.Expression.Value)))
+                [System.IO.File]::WriteAllBytes("$templatePath\Assets\AppIcon.ico", [System.Convert]::FromBase64String(($logo = $defaultAssets.Where({ $_.Item1.Value.Equals('Logo') }).Item2.PipelineElements.Expression.Value)))
+                $configBlock = [System.Management.Automation.ScriptBlock]::Create($ADT.ModuleDefaults.Config.([System.String]::Empty).ToString().Replace($banner, '..\Assets\Banner.Classic.png').Replace($logo, '..\Assets\AppIcon.ico'))
+
+                # Override config values if specified.
+                if ($PSBoundParameters.ContainsKey('Config') -and $Config.Count -gt 0)
```
mjr4077au: Won't need a `$Config.Count` check when the parameter block above has been fixed.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -158,10 +422,97 @@ function New-ADTTemplate
...
+                    $configText = $configBlock.ToString()
+                    $configAst = [System.Management.Automation.Language.Parser]::ParseInput($configText, [ref]$null, [ref]$null)
+                    $configHashtableAst = $configAst.EndBlock.Statements[0].PipelineElements[0].Expression
+                    $configReplacements = [System.Collections.Generic.List[PSCustomObject]]::new()
+                    $configInsertions = @{}
+
+                    # Recursive scriptblock to walk user dictionary against config AST.
+                    $collectReplacements = {
+                        param ([System.Collections.IDictionary]$UserDict, $HashtableAst, [System.String]$Path)
+                        foreach ($key in $UserDict.Keys)
```
mjr4077au: Why not `foreach ($kvp in $UserDict.GetEnumerator())` since you need the value in the `if ($matchingKvp.Count -eq 0)` branch below.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -203,14 +550,151 @@ function New-ADTTemplate
...
+
+                # Process the generated script
                 if ($Version.Equals(4))
                 {
                     $params = @{
                         LiteralPath = "$templatePath\Invoke-AppDeployToolkit.ps1"
                         Encoding = ('utf8', 'utf8BOM')[$PSVersionTable.PSEdition.Equals('Core')]
                     }
-                    Out-File -InputObject (Get-Content @params -Raw).Replace('..\..\..\..\', [System.Management.Automation.Language.NullString]::Value).Replace('2000-12-31', [System.DateTime]::Now.ToString('O').Split('T')[0]) @params -Width ([System.Int16]::MaxValue) -Force
+                    $scriptContent = ((Get-Content @params -Raw) -replace '\r?\n', [System.Environment]::NewLine).Replace('..\..\..\..\', [System.Management.Automation.Language.NullString]::Value).Replace('2000-12-31', [System.DateTime]::Now.ToString('yyyy-MM-dd'))
```
mjr4077au, line 629: What's the `-replace '\r?\n', [System.Environment]::NewLine` stuff for? It'd be better if carriage returns were handled properly throughout the function properly wherever necessary.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
mjr4077au: Don't use the `-split` operator unless you're doing a regex split. The `.Split()` method should be used instead.

mjr4077au (changes requested):
Please have a read through my review comments and go from there. I may want a 2nd review after this first round of review comments have been addressed as there's a lot of code to read here.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
DanGough: It's impossible for us to tell, but I expect there may be some using it on Citrix/Terminal server installs, so it should stay in until we can verify otherwise.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
DanGough: I'm more amenable to dropping this one than TerminalServerMode, but again without some kind of survey/telemetry, it's impossible to tell. I say we leave it in for now until we can get some feedback on this from the community.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: I had some trouble with this! In PS5.1, .Split accepts a char array, so it thinks the \r\n are separate delimiters.

This can be overcome with casting to string and using a different method overload:

image

PS7 doesn't need this workaround though:

image

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
DanGough: I'd prefer it be left in - it's always been a common issue in the past where people stumbled when trying to create user context deployments, and we had to point them to the config to make a change. Now that setting is no longer in the config, this is the only place we can surface it - otherwise it's just buried in the docs.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
mjr4077au: That's not how you use the `.Split()` method, see here for an example: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/37b977147f5bf72d054121c10ea82964ab8bbd0f/src/PSAppDeployToolkit.Build/Private/Show-ADTModuleInitArtwork.ps1#L30

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
DanGough: I've just removed all double linebreaks to keep things shorter.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
mjr4077au: But if it's not defined at all, it'll default to false as it's a switch, meaning it'll just work for people, which in my opinion is the way it should.

Please also don't resolve comments until I add the raiser of the comment consider it resolved. This one needs further discussion and so did another one.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
DanGough: I thought it's not very clear what the purpose of this is otherwise, removing it might lead to questions about how to import extension modules.

src/PSAppDeployToolkit/opt/Frontend/v4/Invoke-AppDeployToolkit.ps1
DanGough: They may look messy, but they are useful if users want their templates to surface all execution errors to the user by default.

My preference would be to keep the 2nd one in there but with the actual code rather than a comment...

If we do keep them though, it's a good idea to add `-Force` to these also.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: Yeah I guess that's simpler, I was aiming for something nice and readable via New-Timespan but this is just as good.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: I have no explanation, but current function in main uses one approach then the other, I guess I unified them the wrong way around - now unified to both use ThrowTerminatingError.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: This particular section is intended to replace an existing value, but I've changed it to use .Add() on the other side of the if block.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: Regex class was used due to it having an embedded scriptblock that replaces tabs with spaces, but I've managed to remove the extra -Replace operation that normalized line endings since we can split on both line ending types instead. It's a bit more readable now, plus there's a comment added to explain.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: The intention is just to normalize all line endings to be the same in case git is configured without core.autocrlf=true for example.

All tests still pass if this is removed, but in reality performance hit is negligible by adding this in.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: This ties in with a previous comment, but I've stripped out the -Replace operation in favour of the more flexible .Split() approach.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
DanGough: This is the last outstanding thing to look at. However after implementing a first pass it felt like it made the code less readable due to the separation and having to jump back and forth.

I will abstract it a bit more though, as it could be useful for repurposing if anybody requests strings.psd1 changes to be shoehorned in here also.

DanGough:
Will rebase this after merging the other open PR with the nuget updates, this will resolve the build issue on this branch.

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
mjr4077au: Thanks, I misread the intention during the initial review.

mjr4077au (approved):

1eyeITguy:
This looks amazing guys.  Looking forward to digging into this.

PR #2136, opened 2026-04-22
Another smattering of miscellaneous things.

### LINKED ISSUE #2135: [Bug] Invoke-AppDeployToolkit.exe doesn't start PS1 installer when both are renamed to AppName_AppVersion.EXE and PS1 - in version 4.2
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.2

### Describe the bug

We change the Invoke-AppDeployToolkit.exe to AppName_AppVersion.EXE and the same name is having the PS1 installer, which was executed when you just started the AppName_AppVersion.EXE - this was working fine in 4.1.8 of PSADT

In 4.2 which I got from @mjr4077au, the same option doesn't work. We can still run installer using AppName_AppVersion.EXE -file 'AppName_AppVersion.PS1' command.

If we try to use Invoke-AppDeployToolkit.exe renamed from 4.1.8 version and run installer created with 4.2 PSADT, execution of installer, using just the renamed Invoke-AppDeployToolkit.exe to AppName_AppVersion.EXE, works fine.
Is it done by design in 4.2 of PSADT or by mistake???

### Steps to reproduce

1. 
2.
3.

### Environment data

```console
Windows 11 24H2
```

mjr4077au:
# Pull Request

## Description

* Move away from `psmodulecache` as it's dead upstream.
* Factor in the session's idle time when determining the RunAsActiveUser.
* Add some nonsense to shut up CodeQL's "Generic catch clause" stuff.
* Factor out some repeated code within `FILETIMEExtensions` class.
* Don't force the `CreateProcessUsingToken()` pathway for the current user with UIAccess apps if they're not the logged on user.
* Refine when to apply `UseShellExecute` and/or `RunAsInvoker` within `ClientServerUtilities.InvokeClientOperationImpl()`.
* Remove the mostly unnecessary private backing field from `ValidateExtensionAttribute`.
* Improve immutability of the `ValidateExtensionAttribute` setup.
* Update all NuGet packages, again.
* Fix numerous typos within a number of comment-based help sections.
* Fix missing `$PSCmdlet.ShouldProcess()` within `Set-ADTIniFunction`.
* Fix bad variable usage within `New-ADTZipFile` function.
* Add `System.IO.Path.Join()` polyfill and use everywhere to shut CodeQL up.
* Refactor `ServerInstance` to make it follow a more usual disposal pattern.
* Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/2135.

## Type of change

## Checklist


src/PSADT/PSADT/Foundation/RunAsActiveUser.cs
```diff
@@ -34,7 +34,7 @@ public sealed record RunAsActiveUser
             // Determine the account that will be used to execute client/server commands in the user's context.
             // Favour the caller's session if it's found and is currently an active user session on the device.
             return (sessionInfo ??= SessionInfo.Get()).FirstOrDefault(static s => (s.SID == AccountUtilities.CallerSid || s.SessionId == AccountUtilities.CallerSessionId) && s.IsActiveUserSession) is not SessionInfo callerSession
-                ? sessionInfo.Where(static s => s.IsActiveUserSession).OrderByDescending(static s => s.LogonTime).FirstOrDefault()?.ToRunAsActiveUser()
+                ? sessionInfo.Where(static s => s.IsActiveUserSession).OrderByDescending(static s => s.LogonTime).ThenBy(static s => s.IdleTime ?? TimeSpan.MaxValue).FirstOrDefault()?.ToRunAsActiveUser()
```
DanGough: Will this have the desired effect? Seems to me it the idle time should be the primary sort here for this to work?

User A logged on 2 hours ago and is busy
User B logged on 1 hour ago but immediately went idle

Wouldn't both lines result in selecting User B?

mjr4077au: Idle time can't be the primary sort otherwise you'd get pot luck as to who gets decided as it'd literally come down to 1/100th of a nanosecond. This is why the most recent session needs to remain the primary sort.

DanGough: But the logon times are static, adding a secondary sort does nothing unless the top 2 users logged in at exactly the same nanosecond and you need a tiebreaker?

It makes sense to me to purely sort by idle time and select the lowest value. I don't see where the pot luck comes in, if we want to choose the user who last moved their mouse or pressed a key? (e.g. in the example above we'd want choose user A rather than B)

mjr4077au: @DanGough after some rest and review after a long weekend in Australia, you're right, but the idle time issue I bought up was legitimate. Picture the following:

* User A logs on at 2026-04-01 at 06:30:00 and has an idle time of 1,500 ns.
* User B logs on at 2026-04-01 at 06:40:00 and has an idle time of 1,700 ns.

What I was trying to avoid was the user who logged on at 06:30:00 being the selected user, purely because their idle time was a mere 200 ns lower. In reality, sub-1ms difference is not something you'd care about with this operation as that's fluctuation due to anything, therefore you'd ideally want the 06:40:00 user being the selected one.

I've reversed the order of those operators, but have quantised the idle time down to test the idle seconds, rather than at the full precision of the TimeSpan object. This means that the difference between 1,500 ns and 1,700 ns will quantise down to 0, meaning no idle time difference between them, allowing the more recently logged on user to be selected.

DanGough: Sounds good. If we really have 2 active users with mere nanoseconds between them, then the toolkit is not really built to deal with that (yet) and we may as well just use Get-Random to give each a fair chance!

DanGough (approved):

PR #2161, opened 2026-05-07
Rewrite service related functions

michael-hollingsworth:
# Pull Request

## Description

- Rewrite service related functions to act more like the built-in cmdlets
- Fix bug where Start/Stop-ADTServiceAndDependencies wasn't interacting with dependent services unless `-SkipDependentServices` was specified
- Fix bug where `-Force` was passed to `Start-Service`

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] General code cleanup (non-breaking change which improves readability)

## Checklist

- [X] I have made corresponding changes to the documentation


mjr4077au:
As discussed privately, on hold for the time being pending some unit tests to round out the PR.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
```diff
@@ -32,26 +38,35 @@ function Private:Invoke-ADTServiceAndDependencyOperation
...
             return
         }
 
+        $serviceParams = @{
+            WarningAction = [System.Management.Automation.ActionPreference]::Ignore
+        }
+
+        if ($Operation -eq 'Stop')
+        {
+            $serviceParams.Add('Force', $true)
```
mjr4077au: This is too blanket to include. Imagine this is a SQL database and we corrupt an .edb file. Including a `-Force` parameter to put this onto the caller would be alright though.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
```diff
@@ -60,33 +75,30 @@ function Private:Invoke-ADTServiceAndDependencyOperation
...
-    if (($Operation -eq 'Stop') -and ($Service.Status -ne 'Stopped'))
+    if (($Operation -eq 'Stop') -and ($Service.Status -ne [System.ServiceProcess.ServiceControllerStatus]::Stopped))
     {
         # Process all dependent services.
         Invoke-ADTDependentServiceOperation
 
         # Stop the parent service.
         Write-ADTLogEntry -Message "Stopping parent service [$($Service.ServiceName)] with display name [$($Service.DisplayName)]."
-        $Service = $Service | Stop-Service -PassThru -WarningAction Ignore -Force
+        $Service = Stop-Service -InputObject $Service -Force -PassThru -WarningAction Ignore
```
mjr4077au: Same here as well. I know it had `-Force` previously on it, but let's not unless the caller requests it.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -13,24 +13,39 @@ function Get-ADTServiceStartMode
...
 
     .EXAMPLE
-        Get-ADTServiceStartMode -Service (Get-Service -Name 'wuauserv')
+        Get-ADTServiceStartMode -InputObject (Get-Service -Name 'wuauserv')
 
         Retrieves the startup mode of the 'wuauserv' service.
 
+    .EXAMPLE
+        ```PowerShell
+        if ((($service = Test-ADTServiceExists -Name 'ScreenConnect*' -PassThru) | Get-ADTServiceStartMode) -ne 'Automatic') {
```
mjr4077au: Nit, but would prefer our Allman bracing here in examples also.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -65,19 +95,23 @@ function Get-ADTServiceStartMode
 
     process
     {
-        Write-ADTLogEntry -Message "Getting the service [$($Service.Name)] startup mode."
         try
         {
             try
             {
+                $gsParams = @{ $PSCmdlet.ParameterSetName = Get-Variable -Name $PSCmdlet.ParameterSetName -ValueOnly }
+                $service = Get-Service @gsParams | Select-Object -First 1
```
mjr4077au: Doesn't this defeat the point of allowing/permitting wildcards?

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -43,19 +58,34 @@ function Get-ADTServiceStartMode
...
+    [System.Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'DisplayName', Justification = "This parameter is accessed programmatically via the ParameterSet it's within, which PSScriptAnalyzer doesn't understand.")]
     [CmdletBinding()]
     [OutputType([System.String])]
     param
     (
-        [Parameter(Mandatory = $true)]
+        [Parameter(Mandatory = $true, ParameterSetName = 'Name')]
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [PSAppDeployToolkit.Attributes.ValidateUnique()]
+        [SupportsWildcards()]
```
mjr4077au: Should we be supporting wildcards here? The functions aren't set up to accept `string[]` input like `Get-Service` provides, and it returns single string values. Maybe we should only allow absolute service names here.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -13,24 +13,39 @@ function Get-ADTServiceStartMode
     .DESCRIPTION
         The `Get-ADTServiceStartMode` function retrieves the startup mode of a specified service. This function checks the service's start type and adjusts the result if the service is set to 'Automatic (Delayed Start)'.
 
-    .PARAMETER Service
-        Specify the service object to retrieve the startup mode for.
+    .PARAMETER Name
+        Specifies the service name of the services to be retrieved the start mode for. Wildcards are permitted. When more than one service matches the wildcard pattern provided, only the service start mode of the first service is returned
```
mjr4077au: This talks of plural services, but only one service is permitted currently due to the parameter being `string`.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -13,24 +13,39 @@ function Get-ADTServiceStartMode
...
+
+    .PARAMETER DisplayName
+        Specifies the display name of the service to retrieve the service start mode for. Wildcards are permitted. When more than one service matches the wildcard pattern provided, only the service start mode of the first service is returned
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -43,19 +58,34 @@ function Get-ADTServiceStartMode
...
+        [Parameter(Mandatory = $true, ParameterSetName = 'Name')]
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [PSAppDeployToolkit.Attributes.ValidateUnique()]
+        [SupportsWildcards()]
+        [System.String]$Name,
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'DisplayName')]
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [PSAppDeployToolkit.Attributes.ValidateUnique()]
+        [SupportsWildcards()]
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -13,21 +13,35 @@ function Set-ADTServiceStartMode
     .DESCRIPTION
         The `Set-ADTServiceStartMode` function sets the service startup mode. This function allows you to configure the startup mode of a specified service. The startup modes available are: Automatic, Automatic (Delayed Start), Manual, Disabled, Boot, and System.
 
-    .PARAMETER Service
-        Specify the name of the service.
+    .PARAMETER Name
+        Specifies the service names of the services to set the start mode for. Wildcards are permitted.
```
mjr4077au: Maybe use "name(s)" here to indicate singular names are acceptable.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -13,21 +13,35 @@ function Set-ADTServiceStartMode
...
+
+    .PARAMETER DisplayName
+        Specifies the display names of services to set the start mode for. Wildcards are permitted.
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -13,21 +13,35 @@ function Set-ADTServiceStartMode
...
+
+    .PARAMETER InputObject
+        Specifies ServiceController objects representing the services to be started.
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -13,21 +13,35 @@ function Set-ADTServiceStartMode
...
     .OUTPUTS
         None
 
-        This function does not return any output.
+        By default, this function does not return any output.
+
+    .OUTPUTS
+        System.ServiceProcess.ServiceController
+
+        When the `-PassThru` parameter is provided, this function returns a ServiceController object representing the service that was started.
```
mjr4077au: Probably should have backticks around the `ServiceController` part.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -81,33 +112,70 @@ function Set-ADTServiceStartMode
...
+                            throw (New-ADTErrorRecord @naerParams)
+                        }
+                        catch
+                        {
+                            Write-Error -ErrorRecord $_
+                        }
+                    }
+                    catch
+                    {
+                        Invoke-ADTFunctionErrorHandler -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState -ErrorRecord $_ -ErrorAction SilentlyContinue
```
mjr4077au, line 202: Nothing to action here for now, but the way we're blanketly applying `-ErrorAction SilentlyContinue` on looped items like this isn't right. There's a better setup in funcs like `Remove-ADTFont` but its still not right either. Once this is merged, I'll circle back around and will address this.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -14,24 +14,27 @@ function Start-ADTServiceAndDependencies
         The `Start-ADTServiceAndDependencies` function starts a specified Windows service and its dependencies. It provides options to skip starting dependent services, wait for a service to get out of a pending state, and return the service object.
 
     .PARAMETER Name
-        Specify the name of the service.
+        Specifies the service names of services to be stopped. Wildcards are permitted.
```
mjr4077au: Maybe use "name(s)" here to indicate singular names are acceptable.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -14,24 +14,27 @@ function Start-ADTServiceAndDependencies
...
+
+    .PARAMETER DisplayName
+        Specifies the display names of services to be stopped. Wildcards are permitted.
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -14,24 +14,27 @@ function Start-ADTServiceAndDependencies
...
 
     .PARAMETER InputObject
-        A ServiceController object to start.
+        Specifies ServiceController objects representing the services to be started.
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -14,24 +14,27 @@ function Start-ADTServiceAndDependencies
...
     .PARAMETER SkipDependentServices
-        Choose to skip checking for and starting dependent services.
+        Specifies whether to skip checking for and starting dependent services.
 
     .PARAMETER PendingStatusWait
         The amount of time to wait for a service to get out of a pending state before continuing. Default is 60 seconds.
 
     .PARAMETER PassThru
-        Return the System.ServiceProcess.ServiceController service object.
+        Return the ServiceController service object.
```
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -13,21 +13,35 @@ function Set-ADTServiceStartMode
...
+
+    .PARAMETER PassThru
+        Returns the ServiceController service object.
 
     .INPUTS
-        None
+        System.ServiceProcess.ServiceController
 
-        You cannot pipe objects to this function.
+        You can pipe ServiceController objects to this function.
```
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -13,21 +13,35 @@ function Set-ADTServiceStartMode
...
+
+    .PARAMETER InputObject
+        Specifies ServiceController objects representing the services to be started.
 
     .PARAMETER StartMode
-        Specify startup mode for the service. Options: Automatic, Automatic (Delayed Start), Manual, Disabled, Boot, System.
+        Specify startup mode for the service. Valid values for this parameter are: `Automatic`, `Automatic (Delayed Start)`, `Manual`, `Disabled`, `Boot`, `System`
+
+    .PARAMETER PassThru
+        Returns the ServiceController service object.
```
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -81,33 +112,70 @@ function Set-ADTServiceStartMode
...
-                Write-ADTLogEntry -Message ($msg = "$msg failed with exit code [$Global:LASTEXITCODE]: $scResult") -Severity Error
-                $naerParams = @{
-                    Exception = [System.Runtime.InteropServices.ExternalException]::new($msg, $Global:LASTEXITCODE)
-                    Category = [System.Management.Automation.ErrorCategory]::InvalidResult
-                    ErrorId = 'ScConfigFailure'
-                    TargetObject = $scResult
-                    RecommendedAction = "Please review the result in this error's TargetObject property and try again."
+                foreach ($service in $services)
+                {
+                    if ($service.StartType -in @('Manual', 'Disabled', 'Boot', 'System') -and $service.StartType -eq $PSBoundParameters.StartMode)
```
mjr4077au: Cache this array in the `begin` block.

mjr4077au: Use braces around the different conditions for reader clarity.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -14,24 +14,27 @@ function Start-ADTServiceAndDependencies
...
 
     .INPUTS
-        None
+        System.ServiceProcess.ServiceController
 
-        You cannot pipe objects to this function.
+        You can pipe a ServiceController object to this function.
```
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +129,29 @@ function Start-ADTServiceAndDependencies
...
             try
             {
-                if ($pipelining)
-                {
-                    $PSBoundParameters.Name = $InputObject.Name
-                }
-                $serviceName = if ($pipelining) { $InputObject.Name } else { $Name }
-                if ($PSCmdlet.ShouldProcess($serviceName, 'Start service and dependencies'))
+                $gsParams = @{ $PSCmdlet.ParameterSetName = Get-Variable -Name $PSCmdlet.ParameterSetName -ValueOnly }
+                $services = Get-Service @gsParams
```
mjr4077au: We should probably only be getting the service again if we're not on the `InputObject` parameter set.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -14,13 +14,16 @@ function Stop-ADTServiceAndDependencies
         The `Stop-ADTServiceAndDependencies` function stops a specified Windows service and its dependencies. It provides options to skip stopping dependent services, wait for a service to get out of a pending state, and return the service object.
 
     .PARAMETER Name
-        Specify the name of the service.
+        Specifies the service names of services to be stopped. Wildcards are permitted.
```
mjr4077au: Maybe use "name(s)" here to indicate singular names are acceptable.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -14,13 +14,16 @@ function Stop-ADTServiceAndDependencies
...
+
+    .PARAMETER DisplayName
+        Specifies the display names of services to be stopped. Wildcards are permitted.
```
mjr4077au: Same here also.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -14,13 +14,16 @@ function Stop-ADTServiceAndDependencies
...
 
     .PARAMETER InputObject
-        A ServiceController object to stop.
+        Specifies ServiceController objects representing the services to be stopped.
```
mjr4077au: Same here also.

mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -13,24 +13,39 @@ function Get-ADTServiceStartMode
...
-    .PARAMETER Service
-        Specify the service object to retrieve the startup mode for.
+    .PARAMETER Name
+        Specifies the service name of the services to be retrieved the start mode for. Wildcards are permitted. When more than one service matches the wildcard pattern provided, only the service start mode of the first service is returned
+
+    .PARAMETER DisplayName
+        Specifies the display name of the service to retrieve the service start mode for. Wildcards are permitted. When more than one service matches the wildcard pattern provided, only the service start mode of the first service is returned
+
+    .PARAMETER InputObject
+        Specify the service object to retrieve the start mode for.
```
mjr4077au: We've used `ServiceController` in other `InputObject` descriptions, but not here?

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -29,9 +32,9 @@ function Stop-ADTServiceAndDependencies
         Return the System.ServiceProcess.ServiceController service object.
```
mjr4077au: You've truncated the name in other `-PassThru` parameters, but not here?

mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -29,9 +32,9 @@ function Stop-ADTServiceAndDependencies
...
 
     .INPUTS
-        None
+        System.ServiceProcess.ServiceController
 
-        You cannot pipe objects to this function.
+        You can pipe a ServiceController object to this function.
```
mjr4077au: Backticks around the type.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +129,29 @@ function Stop-ADTServiceAndDependencies
...
             try
             {
-                if ($pipelining)
-                {
-                    $PSBoundParameters.Name = $InputObject.Name
-                }
-                $serviceName = if ($pipelining) { $InputObject.Name } else { $Name }
-                if ($PSCmdlet.ShouldProcess($serviceName, 'Stop service and dependencies'))
+                $gsParams = @{ $PSCmdlet.ParameterSetName = Get-Variable -Name $PSCmdlet.ParameterSetName -ValueOnly }
+                $services = Get-Service @gsParams
```
mjr4077au: We should probably only be getting the service again if we're not on the InputObject parameter set.

src/PSAppDeployToolkit/Public/Test-ADTServiceExists.ps1
```diff
@@ -14,15 +14,18 @@ function Test-ADTServiceExists
...
 
+    .PARAMETER DisplayName
+        Specifies the display name of the service to test the existence of. This parameter supports wildcards. This parameter is not compaitble with `-UseCIM`.
+
     .PARAMETER UseCIM
         Use CIM/WMI to check for the service. This is useful for compatibility with PSADT v3.x.
 
     .PARAMETER PassThru
-        Return the WMI service object. To see all the properties use: `Test-ADTServiceExists -Name 'spooler' -PassThru | Get-Member`
+        Returns the service object, if one exists. By default, the service object returned is a ServiceController. When the `-UseCIM` parameter is provided, a Win32_Service or Win32_ServiceBase CimInstance is returned.
```
mjr4077au: We should probably only be getting the service again if we're not on the InputObject parameter set.

src/PSAppDeployToolkit/Public/Test-ADTServiceExists.ps1
```diff
@@ -14,15 +14,18 @@ function Test-ADTServiceExists
         The `Test-ADTServiceExists` function checks to see if a service exists. The `-UseCIM` switch can be used in conjunction with `-PassThru` to return WMI objects for PSADT v3.x compatibility, however, this method fails in Windows Sandbox.
 
     .PARAMETER Name
-        Specify the name of the service.
+        Specify the name of the service. When `-UseCIM` is not specified, this parameter supports wildcards.
```
mjr4077au, line 17: Would rather put a deprecation flag around the `-UseCIM` stuff. I doubt anyone's using it.

mjr4077au (changes requested):
Please review the comments and let me know if you've got any questions.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
michael-hollingsworth: I used this as a replacement for:
```PowerShell
if ($PSCmdlet.ParameterSetName -eq 'InputObject')
{
    $InputObject.Refresh()
    $InputObject
}
else
{
...
}
```

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
michael-hollingsworth: I'm not opposed to it but, it would go against the default behavior that's existed since PSADT's original public release: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/3.6.0/Toolkit/AppDeployToolkit/AppDeployToolkitMain.ps1#L8079

If you want, I can add the parameter and write a warning log message about the function's default behavior changing in 4.x.0.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
michael-hollingsworth: Would you prefer a `Get-ADTServiceStartMode` call here or keep the current method for better performance?

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
michael-hollingsworth: Yes and no. It works for services where the full name includes a unique identifier, such as in the ScreenConnect example, where the full name would look something like: `ScreenConnect Client (abcdefghijklmnop)`. I'm personally not a huge fan of this implementation though, so I'll adjust it to filter the output of `Get-Service`.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
mjr4077au: I think that's a good approach if you didn't mind.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
```diff
@@ -11,8 +11,14 @@ function Private:Invoke-ADTServiceAndDependencyOperation
     param
     (
         [Parameter(Mandatory = $true)]
-        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
-        [System.String]$Name,
+        [ValidateScript({
+                if (!$_.ServiceName)
```
mjr4077au: There should be a null check here before trying to access a member of the provided input.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
```diff
@@ -60,33 +75,30 @@ function Private:Invoke-ADTServiceAndDependencyOperation
...
-        Write-ADTLogEntry -Message "Waiting for up to [$($PendingStatusWait.TotalSeconds)] seconds to allow service pending status [$($Service.Status)] to reach desired status [$([System.ServiceProcess.ServiceControllerStatus]$desiredStatus)]."
+        Write-ADTLogEntry -Message "Waiting for up to [$($PendingStatusWait.TotalSeconds)] seconds to allow service pending status [$($Service.Status)] to reach desired status [$desiredStatus]."
         $Service.WaitForStatus($desiredStatus, $PendingStatusWait)
         $Service.Refresh()
     }
 
     # Discover if the service is currently running.
     Write-ADTLogEntry -Message "Service [$($Service.ServiceName)] with display name [$($Service.DisplayName)] has a status of [$($Service.Status)]."
-    if (($Operation -eq 'Stop') -and ($Service.Status -ne 'Stopped'))
+    if (($Operation -eq 'Stop') -and ($Service.Status -ne [System.ServiceProcess.ServiceControllerStatus]::Stopped))
```
mjr4077au: Nit, but when doing equality checks with strongly typed objects, generally the `.Equals()` method is cheaper than PowerShell's `-eq` operator as there's less reflection to go through.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
```diff
@@ -60,33 +75,30 @@ function Private:Invoke-ADTServiceAndDependencyOperation
...
         # Process all dependent services.
         Invoke-ADTDependentServiceOperation
 
         # Stop the parent service.
         Write-ADTLogEntry -Message "Stopping parent service [$($Service.ServiceName)] with display name [$($Service.DisplayName)]."
-        $Service = $Service | Stop-Service -PassThru -WarningAction Ignore -Force
+        $Service = Stop-Service -InputObject $Service -Force -PassThru -WarningAction Ignore
     }
-    elseif (($Operation -eq 'Start') -and ($Service.Status -ne 'Running'))
+    elseif (($Operation -eq 'Start') -and ($Service.Status -ne [System.ServiceProcess.ServiceControllerStatus]::Running))
```
mjr4077au: Nit, but when doing equality checks with strongly typed objects, generally the .Equals() method is cheaper than PowerShell's -eq operator as there's less reflection to go through.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
```diff
@@ -32,26 +38,35 @@ function Private:Invoke-ADTServiceAndDependencyOperation
...
+        if ($Operation -eq 'Stop')
+        {
+            $serviceParams.Add('Force', $true)
+        }
+
         # Action each found dependent service.
         foreach ($dependent in $dependentServices)
         {
-            Write-ADTLogEntry -Message "$(('Starting', 'Stopping')[$Operation -eq 'Start']) dependent service [$($dependent.ServiceName)] with display name [$($dependent.DisplayName)] and a status of [$($dependent.Status)]."
+            Write-ADTLogEntry -Message "$(('Stopping', 'Starting')[$Operation -eq 'Start']) dependent service [$($dependent.ServiceName)] with display name [$($dependent.DisplayName)] and a status of [$($dependent.Status)]."
```
mjr4077au: Just thinking here in retrospect, do we even need any special handling for stopping dependent services? Won't stopping the service in question just act as the lynch pin? Could remove this entire function and inline the code directly into the `elseif (($Operation -eq 'Start') -and ($Service.Status -ne [System.ServiceProcess.ServiceControllerStatus]::Running))` section.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
mjr4077au: Should this be refactored to not be predicated on the `$Service.Status -ne [System.ServiceProcess.ServiceControllerStatus]::Running` part? If the service is started but services depending on it aren't, we probably want to start them?

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -43,19 +59,30 @@ function Get-ADTServiceStartMode
...
+        [System.String]$Name,
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'DisplayName')]
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [System.String]$DisplayName,
+
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true, ParameterSetName = 'InputObject')]
         [ValidateScript({
-                if (!$_.Name)
+                if (!$_.ServiceName)
```
mjr4077au, line 92: There should be a null check here before trying to access a member of the provided input.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -65,19 +92,34 @@ function Get-ADTServiceStartMode
 
     process
     {
-        Write-ADTLogEntry -Message "Getting the service [$($Service.Name)] startup mode."
         try
         {
             try
             {
+                if ($PSCmdlet.ParameterSetName -eq 'InputObject')
```
mjr4077au: Replace this entire if/else setup with something like: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/1d09f57fe069f188e0ded8bb7ba17c1447335fbe/src/PSAppDeployToolkit/Public/Get-ADTExecutableInfo.ps1#L89-L98 The problem with the setup here is that in the else part, it's likely the most use pathway so it should be favoured. Further to this, calling `Get-Service` and filtering after the fact is expensive. By adopting a setup like the above with a splat hashtable, we can get the params right for `Get-Service` ahead of time.

The above example could be further cheapened to something like `$gciParams = @{ $PSCmdlet.ParameterSetName = $PSBoundParameters.($PSCmdlet.ParameterSetName) }` to avoid the Get-Variable call.

If the setup here was to avoid the wildcard issue, just set the value up by doing something like `$gsParams = @{ $PSCmdlet.ParameterSetName = [System.Management.Automation.WildcardPattern]::Escape($PSBoundParameters.($PSCmdlet.ParameterSetName)) }`.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -49,21 +73,38 @@ function Set-ADTServiceStartMode
...
+        [Parameter(Mandatory = $true, ParameterSetName = 'DisplayName')]
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [PSAppDeployToolkit.Attributes.ValidateUnique()]
+        [SupportsWildcards()]
+        [System.String[]]$DisplayName,
+
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true, ParameterSetName = 'InputObject')]
         [ValidateScript({
-                if (!$_.Name)
+                if (!$_.ServiceName)
```
mjr4077au, line 102: There should be a null check here before trying to access a member of the provided input.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -13,24 +13,40 @@ function Get-ADTServiceStartMode
...
-        You cannot pipe objects to this function.
+        You can pipe `ServiceController` objects to this function.
 
     .OUTPUTS
         System.String
 
         Returns the startup mode of the specified service.
 
     .EXAMPLE
-        Get-ADTServiceStartMode -Service (Get-Service -Name 'wuauserv')
```
mjr4077au, line 30: Maybe add two examples for the `-Name`/`-DisplayName` paths above this `-InputObject` path? I expect this path to be the least used pathway, so friendly examples above this would be good.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
```diff
@@ -65,19 +92,34 @@ function Get-ADTServiceStartMode
...
+                {
+                    $InputObject.Refresh()
+                    $service = $InputObject
+                }
+                else
+                {
+                    $serviceName = $PSBoundParameters.($PSCmdlet.ParameterSetName)
+                    if (!($service = Get-Service | & { process { if ($_.($PSCmdlet.ParameterSetName) -eq $serviceName) { return $_ } } }))
+                    {
+                        $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName $PSCmdlet.ParameterSetName -ProvidedValue $PSBoundParameters.($PSCmdlet.ParameterSetName) -ExceptionMessage 'The specified service does not exist.'))
```
mjr4077au: Something to keep in mind, but `$PSCmdlet.ThrowTerminatingError()` does not respond to `-ErrorAction` at all. Someone using this function may wish to do something like `if ((Get-ADTServiceStartMode -Name blah -ErrorAction Ignore) -ne 'Running') { ... }` and in this case they probably should be able to. The suggested refinement in the previous comment would address this though as `Get-Service` will just throw if it doesn't exist, then be handled by `Invoke-ADTFunctionErrorHandler`.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -77,37 +118,90 @@ function Set-ADTServiceStartMode
...
-            return
-        }
         try
         {
             try
             {
-                # Set the start up mode using sc.exe. Note: we found that the ChangeStartMode method in the Win32_Service WMI class set services to 'Automatic (Delayed Start)' even when you specified 'Automatic' on Win7, Win8, and Win10.
-                $scResult = & "$([System.Environment]::SystemDirectory)\sc.exe" config $Service.Name start= $StartMode 2>&1
-                if (!$Global:LASTEXITCODE)
+                $services = if ($PSCmdlet.ParameterSetName -eq 'InputObject')
```
mjr4077au: Nit, but as mentioned above, I generally try to favour the most likely scenario in the `if` part of a branch, mostly stemming from my system developer background as it can help with branch prediction. I wouldn't refuse the PR if this wasn't changed, more just an FYI.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -77,37 +118,90 @@ function Set-ADTServiceStartMode
...
+                        continue
+                    }
+
+                    try
+                    {
+                        try
+                        {
+                            # Set the start up mode using sc.exe. Note: we found that the ChangeStartMode method in the Win32_Service WMI class set services to 'Automatic (Delayed Start)' even when you specified 'Automatic' on Win7, Win8, and Win10.
+                            $scResult = & "$([System.Environment]::SystemDirectory)\sc.exe" config $service.ServiceName start= $StartMode 2>&1
+                            if (!$Global:LASTEXITCODE)
```
mjr4077au: I'd have probably reversed this if statement setup so it was `if ($Global:LASTEXITCODE) { ... }` with the thrown ErrorRecord within as it'd avoid two lines of `continue`.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -72,30 +75,42 @@ function Start-ADTServiceAndDependencies
...
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [PSAppDeployToolkit.Attributes.ValidateUnique()]
+        [SupportsWildcards()]
+        [System.String[]]$DisplayName,
 
-        [Parameter(Mandatory = $true, ParameterSetName = 'InputObject')]
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true, ParameterSetName = 'InputObject')]
         [ValidateScript({
-                if (!$_.Name)
+                if (!$_.ServiceName)
```
mjr4077au, line 108: There should be a null check here before trying to access a member of the provided input.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +129,37 @@ function Start-ADTServiceAndDependencies
...
+
+                foreach ($service in $services)
+                {
+                    try
+                    {
+                        try
+                        {
+                            if ($PSCmdlet.ShouldProcess($service.ServiceName, "Start service$(if (!$SkipDependentServices) { ' and dependencies' })"))
+                            {
+                                Invoke-ADTServiceAndDependencyOperation -Service $service -Operation Start @iasadoParams
```
mjr4077au: Should `Invoke-ADTServiceAndDependencyOperation` just accept array input on its parameters to move this loop logic into there? It'd tighten up these `Start-*` and `Stop-` functions.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +129,37 @@ function Stop-ADTServiceAndDependencies
...
+
+                foreach ($service in $services)
+                {
+                    try
+                    {
+                        try
+                        {
+                            if ($PSCmdlet.ShouldProcess($service.ServiceName, "Stop service$(if (!$SkipDependentServices) { ' and dependencies' })"))
+                            {
+                                Invoke-ADTServiceAndDependencyOperation -Service $service -Operation Stop @iasadoParams
```
mjr4077au: Should Invoke-ADTServiceAndDependencyOperation just accept array input on its parameters to move this loop logic into there? It'd tighten up these Start-* and Stop- functions.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -72,30 +75,42 @@ function Stop-ADTServiceAndDependencies
...
+        [PSAppDeployToolkit.Attributes.ValidateNotNullOrWhiteSpace()]
+        [PSAppDeployToolkit.Attributes.ValidateUnique()]
+        [SupportsWildcards()]
+        [System.String[]]$DisplayName,
 
-        [Parameter(Mandatory = $true, ParameterSetName = 'InputObject')]
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true, ParameterSetName = 'InputObject')]
         [ValidateScript({
-                if (!$_.Name)
+                if (!$_.ServiceName)
```
mjr4077au, line 108: There should be a null check here before trying to access a member of the provided input.

mjr4077au (changes requested):
Please review the latest feedback and let me know your thoughts.

src/PSAppDeployToolkit/Public/Get-ADTServiceStartMode.ps1
michael-hollingsworth: It was to avoid passing wildcards to `Get-Service`. Not sure why I didn't go with the WildcardPatter.Escape() method.

src/PSAppDeployToolkit/Private/Invoke-ADTServiceAndDependencyOperation.ps1
michael-hollingsworth: In that case, we could just scrap the entire `Invoke-ADTServiceAndDependencyOperation` function. `Stop-ADTServiceAndDependencies` could just be `Stop-Service -Force` with some additional logging to state which dependent services would be stopped in the process. `Start-ADTServiceAndDependencies` would inherit the dependency operation code from `Invoke-ADTServiceAndDependencyOperation`. Let me know what you think of that.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -77,37 +127,88 @@ function Set-ADTServiceStartMode
...
             try
             {
-                # Set the start up mode using sc.exe. Note: we found that the ChangeStartMode method in the Win32_Service WMI class set services to 'Automatic (Delayed Start)' even when you specified 'Automatic' on Win7, Win8, and Win10.
-                $scResult = & "$([System.Environment]::SystemDirectory)\sc.exe" config $Service.Name start= $StartMode 2>&1
-                if (!$Global:LASTEXITCODE)
+                $services = if ($PSCmdlet.ParameterSetName -ne 'InputObject')
                 {
-                    Write-ADTLogEntry -Message "Successfully set service [$($Service.Name)] startup mode to [$StartMode]."
-                    return
+                    $gsParams = @{ $PSCmdlet.ParameterSetName = Get-Variable -Name $PSCmdlet.ParameterSetName -ValueOnly }
```
mjr4077au: `$gsParams = @{ $PSCmdlet.ParameterSetName = $PSBoundParameters.($PSCmdlet.ParameterSetName) }` would be cheaper here.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -77,37 +127,88 @@ function Set-ADTServiceStartMode
...
-                Write-ADTLogEntry -Message ($msg = "$msg failed with exit code [$Global:LASTEXITCODE]: $scResult") -Severity Error
-                $naerParams = @{
-                    Exception = [System.Runtime.InteropServices.ExternalException]::new($msg, $Global:LASTEXITCODE)
-                    Category = [System.Management.Automation.ErrorCategory]::InvalidResult
-                    ErrorId = 'ScConfigFailure'
-                    TargetObject = $scResult
-                    RecommendedAction = "Please review the result in this error's TargetObject property and try again."
+                foreach ($service in $services)
+                {
+                    if (($service.StartType -in $startTypes) -and ($service.StartType -eq $PSBoundParameters.StartMode))
```
mjr4077au: Why `$PSBoundParameters.StartMode` instead of just `$StartMode`?

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
```diff
@@ -77,37 +127,88 @@ function Set-ADTServiceStartMode
                 'Manual' { 'Demand'; break }
                 default { $_; break }
             })
+
+        # All the start types except for 'Automatic' since we aren't validating whether 'Automatic' means 'Automatic' or 'Automatic (Delayed Start)'
+        $startTypes = @(
```
mjr4077au: Why not just `!$service.StartType.Equals([System.ServiceProcess.ServiceStartMode]::Automatic)` instead of this lookup table?

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +141,85 @@ function Start-ADTServiceAndDependencies
         {
             try
             {
-                if ($pipelining)
+                $services = if ($PSCmdlet.ParameterSetName -ne 'InputObject')
                 {
-                    $PSBoundParameters.Name = $InputObject.Name
+                    $gsParams = @{ $PSCmdlet.ParameterSetName = Get-Variable -Name $PSCmdlet.ParameterSetName -ValueOnly }
```
mjr4077au: `$gsParams = @{ $PSCmdlet.ParameterSetName = $PSBoundParameters.($PSCmdlet.ParameterSetName) }` would be cheaper here.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +141,85 @@ function Start-ADTServiceAndDependencies
...
+                            if (($desiredStatus = $desiredStatusLookupTable[$service.Status]))
+                            {
+                                Write-ADTLogEntry -Message "Waiting for up to [$($PendingStatusWait.TotalSeconds)] seconds to allow service pending status [$($service.Status)] to reach desired status [$desiredStatus]."
+                                $service.WaitForStatus($desiredStatus, $PendingStatusWait)
+                                $service.Refresh()
+                            }
+
+                            if (!$SkipDependentServices)
+                            {
+                                $dependentServices = $service.DependentServices | & { process { if (!$_.Status.Equals([System.ServiceProcess.ServiceControllerStatus]::Running)) { return $_ } } }
```
mjr4077au: Do `$dependentServices = if (!$SkipDependentServices) { ... }` as this variable assignment in the if statement won't exist under some circumstances, or will point to old data on the next loop if the branch isn't hit.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +141,85 @@ function Start-ADTServiceAndDependencies
...
+                            }
+
+                            if (!$service.Status.Equals([System.ServiceProcess.ServiceControllerStatus]::Running))
```
mjr4077au: Because this serves as a guard clause, generally I'd flip the condition as its less natural to short circuit in an else statement.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +158,66 @@ function Stop-ADTServiceAndDependencies
         {
             try
             {
-                if ($pipelining)
+                $services = if ($PSCmdlet.ParameterSetName -ne 'InputObject')
                 {
-                    $PSBoundParameters.Name = $InputObject.Name
+                    $gsParams = @{ $PSCmdlet.ParameterSetName = Get-Variable -Name $PSCmdlet.ParameterSetName -ValueOnly }
```
mjr4077au: `$gsParams = @{ $PSCmdlet.ParameterSetName = $PSBoundParameters.($PSCmdlet.ParameterSetName) }` would be cheaper here.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +141,85 @@ function Start-ADTServiceAndDependencies
...
+                                continue
+                            }
+
+                            if (!$service.Status.Equals([System.ServiceProcess.ServiceControllerStatus]::Running))
+                            {
+                                Write-ADTLogEntry -Message "Starting parent service [$($service.ServiceName)] with display name [$($service.DisplayName)]."
+                                Start-Service -InputObject $service -PassThru:$PassThru -WarningAction Ignore
+                            }
+
+                            foreach ($dependent in $dependentServices)
```
mjr4077au, line 195: In the `Stop-` function, we're doing `Write-ADTLogEntry -Message "The following dependent service(s) [$($dependentServiceNames -join ', ')] will be stopped by this operation."`.

My question is, do we need to be so verbose here? Is it enough to just do `Start-Service -Name $dependentServices.ServiceName` and move on or do you think the added verbosity serves benefit?

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -8,30 +8,36 @@ function Stop-ADTServiceAndDependencies
...
+    .PARAMETER DisplayName
+        Specifies the display name(s) of services to be stopped. Wildcards are permitted.
 
     .PARAMETER InputObject
-        A ServiceController object to stop.
+        Specifies `ServiceController` object(s) representing the services to be stopped.
 
     .PARAMETER SkipDependentServices
-        Choose to skip checking for and stopping dependent services.
+        Specifies whether to skip checking for and stopping dependent services.
```
mjr4077au: We should flag that this parameter is obsolete here also, just like we're doing with `-IgnoreExitCodes` on the `Start-ADT*Process*` functions.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -72,42 +83,72 @@ function Stop-ADTServiceAndDependencies
...
+
         [Parameter(Mandatory = $false)]
         [System.Management.Automation.SwitchParameter]$PassThru
     )
 
     begin
     {
         Initialize-ADTFunction -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
-        if ($pipelining = $PSCmdlet.ParameterSetName.Equals('InputObject'))
+        $desiredStatusLookupTable = @{
```
mjr4077au: Worth moving this to `ImportsLast.ps1` via `$Script:ADT.ServiceStatusMap` or something? Its not that its expensive to define, it's just that the same thing is defined twice across like functions.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
```diff
@@ -105,9 +127,11 @@ function Start-ADTServiceAndDependencies
     begin
     {
         Initialize-ADTFunction -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
-        if ($pipelining = $PSCmdlet.ParameterSetName.Equals('InputObject'))
-        {
-            $null = $PSBoundParameters.Remove('InputObject')
+        $desiredStatusLookupTable = @{
```
mjr4077au: Worth moving this to `ImportsLast.ps1` via `$Script:ADT.ServiceStatusMap` or something? Its not that its expensive to define, it's just that the same thing is defined twice across like functions.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -72,42 +83,72 @@ function Stop-ADTServiceAndDependencies
...
+
+        if ($SkipDependentServices)
         {
-            $null = $PSBoundParameters.Remove('InputObject')
+            Write-ADTLogEntry -Message 'The [-SkipDependentServices] parameter is no longer required and will be removed in PSAppDeployToolkit 4.3.0.' -Severity Warning
+        }
+
+        if (!$Force)
+        {
+            Write-ADTLogEntry -Message 'Starting in PSAppDeployToolkit 4.3.0, the [-Force] parameter will be required to stop services when services that depend on it are running.' -Severity Warning
```
mjr4077au: I feel I've made a mistake and possibly misguided you during this PR... Is the `-Force` parameter on `Stop-Service` necessary to stop dependent services as part of stopping the requested service? I thought it was just to terminate a service that wouldn't stop otherwise.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -117,14 +158,66 @@ function Stop-ADTServiceAndDependencies
...
+                        try
+                        {
+                            if (($desiredStatus = $desiredStatusLookupTable[$service.Status]))
+                            {
+                                Write-ADTLogEntry -Message "Waiting for up to [$($PendingStatusWait.TotalSeconds)] seconds to allow service pending status [$($service.Status)] to reach desired status [$desiredStatus]."
+                                $service.WaitForStatus($desiredStatus, $PendingStatusWait)
+                                $service.Refresh()
+                            }
+
+                            if (!$service.Status.Equals([System.ServiceProcess.ServiceControllerStatus]::Stopped))
```
mjr4077au: Because this serves as a guard clause, generally I'd flip the condition as its less natural to short circuit in an else statement.

mjr4077au (changes requested):
Only some minor feedback. My main concern is around the `-Force` parameter on `Stop-ADTServiceAndDependencies`. I didn't realise it was required for stopping a service if it had dependencies, rather it was to kill a service that wouldn't stop otherwise. If that's the case, the `-Force` parameter can probably be removed, sorry...

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
michael-hollingsworth: It doesn't necessarily have to be more verbose. My thought process is that doing it this way results in a log entry for every action performed by the function. That's why is shortened it in the `Stop-` function; It isn't actually calling `Stop-Process` for every dependent, therefore they can all be balled up into one log message. It also keeps the logging similar to how it was before this PR.

mjr4077au: Too easy, let's leave this one as it is then.

src/PSAppDeployToolkit/Public/Start-ADTServiceAndDependencies.ps1
mjr4077au: Probably more like `New-Variable -Name ServiceStatusTable -Option Constant -Value ...`, keeping in line with some of the other tables there.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
mjr4077au: Probably more like `New-Variable -Name ServiceStatusTable -Option Constant -Value ...`, keeping in line with some of the other tables there.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
michael-hollingsworth: From what I can tell that's the case: Forces the cmdlet to stop a service even if that service has dependent services.

I'll drop the param.

src/PSAppDeployToolkit/Public/Set-ADTServiceStartMode.ps1
michael-hollingsworth: The `$StartMode` variable is overwritten in the begin block so we can't trust it

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
```diff
@@ -72,30 +80,45 @@ function Stop-ADTServiceAndDependencies
...
-                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Service -ProvidedValue $_ -ExceptionMessage 'The specified service does not exist.'))
+                    $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName InputObject -ProvidedValue $_ -ExceptionMessage 'The specified service does not exist.'))
                 }
                 return !!$_
             })]
-        [System.ServiceProcess.ServiceController]$InputObject,
+        [System.ServiceProcess.ServiceController[]]$InputObject,
 
         [Parameter(Mandatory = $false)]
+        [System.Obsolete("This parameter is not supported and will be removed in PSAppDeployToolkit 4.3.0.")]
```
mjr4077au: Should we be obsoleting this parameter? I thought we'd use it to determine whether to not apply `-Force`, etc.

mjr4077au (changes requested):
Only one comment to review/discuss. Let me know your thoughts.

src/PSAppDeployToolkit/Public/Stop-ADTServiceAndDependencies.ps1
michael-hollingsworth: The default non-terminating exception passed by `Stop-Service` informs the user to use the `-Force` parameter to stop services with running dependents. To prevent confusion, I added a log message stating that the service was skipped with an early return.
https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs#L1026-L1046

mjr4077au (approved):

Issue #180, opened 2016-05-17
$envOSVersion not reliable in editors like ISE and PowerGUI


That-Annoying-Guy:
If you run the following line from AppDeployToolkitMain.ps1 in ISE or PowerGUI, you get different versions than straight PowerShell.exe:

```
[version]$envOSVersion = [Environment]::OSVersion.Version
```

This more noticeable in Windows 8.1 where it returns Windows 8's version number!

Meanwhile it looks like  **.OSVersion is deprecated** 
We should use something else because in future versions of .Net it may be removed and no longer supported.  
http://msdn.microsoft.com/en-us/library/system.environment.osversion(v=vs.100).aspx
(Notice it doesn't come up for .Net 4.5 and above) 

Workaround:

```
[version]$envOSVersion = ([Regex]'(\d+\.){2}\d+').Match( (get-item (Join-Path ([Environment]::SystemDirectory) 'ntdll.dll')).VersionInfo.FileVersion ).Value
```

Source: http://powershell.com/cs/media/p/41298.aspx

amotaz:
I was able to reproduce this issue in windows 8 and I added a fix. Please test to confirm. Thank you for the find.

That-Annoying-Guy:
It works in Win7 fully patched and win7 vanilla unpatched.

Thank you!

amotaz:
thank you for the confirmation. I was able to test on Windows 8.1 and Windows 10 as well, but I welcome any other results.

That-Annoying-Guy:
FYI: fix not mentioned in CHANGELOG.txt

That-Annoying-Guy:
FYI: This change has been undone in change labelled "Updated how OS version number are handled"

 **.OSVersion is deprecated**

PR #2166, opened 2026-05-14
Even more minor fixes/changes/improvements for 4.2.0.

mjr4077au:
# Pull Request

## Description

* Make sure the BlockExecution stuff always uses the non-UIAccess launcher.
* Fix bad length check in `Block-ADTAppExecution`.
* Tighten up some null checking in `LogUtilities.WriteLogEntry()`.
* Do extension testing in `ValidateExtensionAttribute.ValidateElement()`.
* Validate input length in `ValidateExtensionAttribute` ctor.
* Include hidden/system files in `Mount-ADTWimFile` when testing a directory.
* Refactor the client/server `RemoveArgumentsDictionaryStorage` setup.
* Use `WindowTools.BringWindowToFront()` to try and help make dialogs show on top.
* Improve `AssemblyPermissions.Remediate()` for non-admin calling users.
* Move the Fluent balloon tip suppression into the Show/Close-ADTInstallationProgress functions.
* Swap classic/fluent  dialog style default positions in `Block-ADTAppExecution` if not specified.
* Myriad of code changes/cleanups/improvements, mostly for CodeQL.

## Type of change

## Checklist

## How Has This Been Tested?

Tested default script template on my machine under admin/non-admin use cases. Also performed testing with @ASDF-PS as part of some of these `-BlockExecution` refinements.


src/PSAppDeployToolkit/Public/Mount-ADTWimFile.ps1
```diff
@@ -114,7 +114,7 @@ function Mount-ADTWimFile
                 {
                     $PSCmdlet.ThrowTerminatingError((New-ADTValidateScriptErrorRecord -ParameterName Path -ProvidedValue $_ -ExceptionMessage 'The specified mount path has a pre-existing WIM mounted.'))
                 }
-                if (Get-ChildItem -LiteralPath $_ -ErrorAction Ignore)
+                if (Get-ChildItem -LiteralPath $_ -Force)
```
DanGough: We still need the `-ErrorAction Ignore` here I think - I can no longer mount a WIM file to a directory that does not pre-exist (I could before this change):

> Mount-ADTWimFile : Cannot validate argument on parameter 'Path'. The running command stopped because the preference
> variable "ErrorActionPreference" or common parameter is set to Stop: Cannot find path 'C:\Temp\Testy' because it does
> not exist.

src/PSAppDeployToolkit/Public/Block-ADTAppExecution.ps1
```diff
@@ -152,7 +152,11 @@ function Block-ADTAppExecution
                     ButtonRightText = [PSADT.UserInterface.BlockExecution]::ButtonText
                     Icon = [PSADT.UserInterface.DialogSystemIcon]::Warning
                 }
-                if ($PSBoundParameters.ContainsKey('WindowLocation'))
+                if (!$PSBoundParameters.ContainsKey('WindowLocation'))
```
DanGough, line 155: This doesn't take into account the fact the user could have overriden the default dialog position. However it would be a lot of extra work to track all open dialog positions including async ones and choose a location that does not overlap.

(so not asking for changes here, merely a comment)

mjr4077au: It does, that's in the `else` part of this branch.

src/PSAppDeployToolkit/Public/Mount-ADTWimFile.ps1
mjr4077au: This entire check should be removed, actually, as we have a `-Force` parameter to handle this but we can't test here. That's all handled on Line #180 though.

DanGough (changes requested):

src/PSAppDeployToolkit/Public/Block-ADTAppExecution.ps1
DanGough: I meant they could have overridden the location of some other clashing dialog such as Show-ADTInstallationProgress.

src/PSAppDeployToolkit/Public/Mount-ADTWimFile.ps1
mjr4077au: This should now be all fixed up.

DanGough (approved):

PR #2192, opened 2026-06-19
Add `-Clear` parameter to `Set-ADTShortcut` function.

### LINKED ISSUE #2177: [Feature] Allow Set-ADTShortcut to clear shortcut command line arguments

Set-ADTShortcut has the -Arguments option to modify the command line arguments for a shortcut; however, it does not allow an empty string to be provided as the value.

Developers have indicated that this is a safety feature ("having checks like this to validate the input typically leads to your code being better or more robust by throwing early rather than nuking a shortcut on 10,000 workstations before you realise you've got a problem").

Developers also suggested that a new option called something like -ClearArguments could be added to provide this functionality with less risk. Were this to be implemented, Set-ADTShortcut should probably throw an error if both the new option and -Arguments are specified together.

mjr4077au:
# Pull Request

## Description

It's been requested that the ability to clear a shortcut's properties be added to `Set-ADTShortcut`. From the backend's perspective, this is performed by sending `null` through to the `Arguments` property of a `ShellLinkFile` object.

Via this PR, I've implemented via via a `-ClearArguments` switch parameter. It's been suggested that `-Arguments` should be made to accept null, but there are several issues with this:

* The `Set-ADTShortcut` command has been around for [over 6 years](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/c90873c1ae8c5b3e4f6405cf4b4f910c4901c2fe), and has always properly validated its input.
* Due to its longevity, callers will have come to expect that the function should properly validate it's inputs, such as referencing a missing key in a hashtable as a potential parameter value.
* I strongly feel any/all parameters that accept input should validate the input is not null or empty to ensure an integrator does not obliterate the arguments out of a shortcut. Mistakes happen, we're all human.
* The benefit of using a toolkit like PSAppDeployToolkit is that we provide such safeties and well thought out processes, in addition to our logging.
* On a technical level, it's not possible to pass `$null` through to a `[System.String]` variable in PowerShell. While the parameter could be set up with an `[AllowNull()]` decoration, the only way to pass a true null would be for the caller to specify `[NullString]::Value` as a value, with all other values being interpreted as an empty string. This is not intuitive and beyond the technical realm of most of our users who probably don't even know this static singleton class exists.
* Attempting to weaken the public contract further by making the parameter `[System.Object]` is also worse as while it would allow `$null` to be specified, it'll require more advanced checks, make intellisense via Visual Studio code less intuitive, and will compromise the help output in PowerShell and our website.

A convention I'd support as an alternative would be a `Clear-ADTShortcutProperty` function. This would be analogous with [`Clear-ItemProperty`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/clear-itemproperty?view=powershell-7.6) from PowerShell's builtin functionality and specifically exists to allow a strong contract on `Set-ItemProperty` while still providing the ability to specifically clear an item property if required.

Fixes: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/2177.

## Type of change

## Checklist


michael-hollingsworth:
I'd prefer a `Set-ADTShortcut -Clear` parameter, kind of like what `Set-ADUser` has or the `Clear-ADTShortcutProperty` function you proposed in the description of this PR. That way, it could be used to clear the arguments, description, hotkey, etc. of a shortcut.

DanGough:
My major bugbear with `-ClearArguments` is that it only exposes one of the properties you'd potentially want to clear, and adding `-ClearDescription`, `-ClearHotKey`, etc would quickly become a mess.

We allow null inputs in Set-ADTIniValue and Set-ADTRegistryKey already (e.g. `Set-ADTRegistryKey -LiteralPath 'HKCU\Software\Test' -Name 'Test' -Value $null` works without having to resort to `''` or `[NullString]::Value`), as does Microsoft in Set-ItemProperty, so this would be the simplest approach.

However I think the best solution would be as @michael-hollingsworth suggested, have a single `-Clear` parameter that accepts an array of nullable shortcut properties, in the same vein as `Set-ADUser`:

https://learn.microsoft.com/en-us/powershell/module/activedirectory/set-aduser?view=windowsserver2025-ps#-clear

It would need a check to verify that the user was not trying to both set and clear the same property at the same time, that should result in a throw.

mjr4077au:
Agreed with all parties that, in hindsight, `-ClearArguments` is limiting whereas a generic `-Clear` parameter would be better.

Regarding `Set-ADTRegistryKey` and `Set-ItemProperty`, these are not good reference cases for the reasons mentioned in the initial PR text. As registry values could be a `string`, `int`, `byte[]`, or even `long`, we, and PowerShell have the `-Value` parameter typed as `[object]`, which matches the underlying `Microsoft.Win32.Registry` implementation. A `$null` can be sent to an `[object]` parameter purely because of the parameter type, not because its intended. The intention is `Clear-ItemProperty` be used which is why it exists.

Because `Set-ADTShortcut -Arguments` is strongly typed as a `[string]`, a `$null` will be coerced into `''` by PowerShell, with only `[NullString]::Value` being able to be sent if null is to be truly provided. I don't want to weaken the callsite by weaking the typing on the function, and I don't want callers to have to specify `[NullString]::Value` to clear something both because it's obtuse to the caller, and the aforementioned risks of allowing null input.

I've updated this PR to use a generic `-Clear` parameter instead which should hopefully address everything.

DanGough (changes requested):
For .lnk shortcuts, the following errors are seen when trying to clear certain properties:

**TargetPath:**
`Value does not fall within the expected range.`

**IconLocation:**
`Cannot convert the "" value of type "System.Management.Automation.Language.NullString" to type "System.IO.FileInfo".`

**WindowStyle:**
`Cannot convert null to type "PSADT.ShortcutManagement.ShortcutWindowStyle" due to enumeration values that are not valid. Specify one of the following enumeration values and try again. The possible enumeration values are "Hidden,Normal,Minimized,Maximized,NormalNoActivate,NormalNoRestore,MinimizedActivateRecent,MinimizedNoActivate,NormalNoRestoreNoActivate,Restore,ProcessDefault,ForceMinimized"`

For .url shortcuts, only **TargetPath** has an issue:
`Cannot convert the "" value of type "System.Management.Automation.Language.NullString" to type "System.Uri`

I suggest we:
- Remove TargetPath from the ValidateSet, this is the core of the shortcut and doesn't make sense to allow it to be empty.
- Either remove WindowStyle also, or rig it up so that it sets it back to the default of Normal for .lnk shortcuts.
- IconLocation - it's valid to set this as empty for .url shortcuts... so need to find a solution for .lnk.

DanGough (approved):

PR #2218, opened 2026-07-01
Read only fixes

DanGough:
# Pull Request

## Description

2 fixes:

- When creating a template, module files are marked as read only. When using this read-only module to create a template, that ends up with the user-editable Invoke-AppDeployToolkit.ps1 (and also the PSAppDeployToolkit.Extensions files) being marked as read-only, which causes New-ADTTemplate to throw an error. Read-only attributes are now removed from all *.ps*1 files in the template before the module is copied in.

- Existing code for marking the module files read-only was clobbering existing attributes and setting read-only. Now it preserves other existing attributes.

## Type of change

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas

## How Has This Been Tested?

Tested by creating a template then importing that module in a fresh session to create another template - which previously threw an error.
Also verified that other attributes (e.g. hidden/compress/encrypt) are preserved.

### DIFF
--- a/src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
+++ b/src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
@@ -590,10 +590,16 @@ function New-ADTTemplate
                 }
                 Export-ADTScriptBlockToFile -ScriptBlock $Script:ADT.ModuleDefaults.Strings.([System.String]::Empty) -LiteralPath "$templatePath\Strings\strings.psd1"
 
-                # Remove any digital signatures from the ps*1 files.
+                # Ensure all editable ps*1 files are not read-only and remove any digital signatures.
                 Get-ChildItem -LiteralPath $templatePath -File -Filter *.ps*1 -Recurse | & {
                     process
                     {
+                        $attributes = [System.IO.File]::GetAttributes($_.FullName)
+                        if (($attributes -band [System.IO.FileAttributes]::ReadOnly) -ne 0)
+                        {
+                            [System.IO.File]::SetAttributes($_.FullName, ($attributes -band (-bnot [System.IO.FileAttributes]::ReadOnly)))
+                        }
+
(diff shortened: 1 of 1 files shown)


mjr4077au (approved):

src/PSAppDeployToolkit/Public/New-ADTTemplate.ps1
```diff
@@ -590,10 +590,16 @@ function New-ADTTemplate
                 }
                 Export-ADTScriptBlockToFile -ScriptBlock $Script:ADT.ModuleDefaults.Strings.([System.String]::Empty) -LiteralPath "$templatePath\Strings\strings.psd1"
 
-                # Remove any digital signatures from the ps*1 files.
+                # Ensure all editable ps*1 files are not read-only and remove any digital signatures.
                 Get-ChildItem -LiteralPath $templatePath -File -Filter *.ps*1 -Recurse | & {
                     process
                     {
+                        $attributes = [System.IO.File]::GetAttributes($_.FullName)
```
michael-hollingsworth, line 597: Is there any particular reason why you did:
```PowerShell
$attributes = [System.IO.File]::GetAttributes($_.FullName)
if (($attributes -band [System.IO.FileAttributes]::ReadOnly) -ne 0)
{
    [System.IO.File]::SetAttributes($_.FullName, ($attributes -band (-bnot [System.IO.FileAttributes]::ReadOnly)))
}
```
instead of:
```PowerShell
if ($_.Attributes -band [System.IO.FileAttributes]::ReadOnly) -ne 0)
{
    $_.Attributes = $_.Attributes -band (-bnot [System.IO.FileAttributes]::ReadOnly)
}
```

DanGough: Nothing other than stupidity/ignorance! The function had `[System.IO.File]::SetAttributes()` previously so I re-used that pattern without realising there was already an Attributes property available.

I can update this later, but also happy for you to do so since you pointed it out.

PR #2236, opened 2026-07-04
Add -PerUser support to Register-ADTDll, Unregister-ADTDll and Invoke-ADTRegSvr32

pl4nty:
# Pull Request

## Description

Adds a -PerUser switch for DLL registration to run in the user's context, including the logged-on user if running as SYSTEM.

I was [fixing the Teams Outlook addin](https://learn.microsoft.com/en-us/troubleshoot/microsoftteams/meetings/resolve-teams-meeting-add-in-issues#reregister-the-teams-addin-loader), but found I couldn't use `Register-ADTDll` because I was running in SYSTEM context.

Happy to make this a feature request instead if that's preferred. I also wasn't sure about the structure, `Register-ADTDllAsUser` felt like overkill but I couldn't find any other cmdlets with a user-context switch. I went with `-PerUser` to avoid ambiguity with the `-User [Identifier]` parameter used elsewhere. `-AsUser` might be better though?

## Type of change

## Checklist

I assume the docstring changes will update the docs automatically.

## How Has This Been Tested?

Used the cmdlets to unregister and register `%LocalAppData%\Microsoft\TeamsMeetingAdd-in\1.26.16801\x64\Microsoft.Teams.AddinLoader.dll`

### DIFF
--- a/src/PSAppDeployToolkit/Public/Invoke-ADTRegSvr32.ps1
+++ b/src/PSAppDeployToolkit/Public/Invoke-ADTRegSvr32.ps1
@@ -19,7 +19,7 @@ function Invoke-ADTRegSvr32
     .PARAMETER Action
         Specify whether to register or unregister the DLL.
 
-    .PARAMETER PerUser
+    .PARAMETER AsUser
         Specifies that the DLL should be registered for the current user only by calling its DllInstall entry point with the 'user' argument (regsvr32.exe /n /i:user). If this function is running under the SYSTEM account, regsvr32.exe is executed in the context of the currently logged on user. Note that the DLL must support per-user registration via a DllInstall export for this to work.
 
     .INPUTS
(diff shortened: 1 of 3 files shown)


src/PSAppDeployToolkit/Public/Invoke-ADTRegSvr32.ps1
```diff
@@ -153,8 +164,27 @@ function Invoke-ADTRegSvr32
                     throw (New-ADTErrorRecord @naerParams)
                 }
 
-                # Register the DLL file and measure the success.
-                Start-ADTProcess -FilePath $RegSvr32Path -ArgumentList $ActionParameters -CreateNoWindow -SuccessExitCodes 0
+                # Register the DLL file and measure the success. Per-user registration writes to the
+                # user's HKCU hive, so when running as SYSTEM, execute as the logged on user instead.
+                if ($PerUser -and [PSADT.AccountManagement.AccountUtilities]::CallerSid.IsWellKnown([System.Security.Principal.WellKnownSidType]::LocalSystemSid))
+                {
+                    if (!(Get-ADTClientServerUser))
```
mjr4077au: `Start-ADTProcessAsUser` will also perform such a check and throw, unless you're doing it here for a more specific exception message?

mjr4077au (dismissed):
All looks pretty good to me. Approved but just had one question before I hit the button.

mjr4077au:
> `-AsUser` might be better though?

Regarding this, I do think so as `-PerUser` implies it's going to cycle through all logged on user sessions.

src/PSAppDeployToolkit/Public/Invoke-ADTRegSvr32.ps1
pl4nty: Ah I wasn't sure, makes sense. I've removed the redundant guards then

mjr4077au (approved):

PR #2264, opened 2026-07-30
Improvements to Copy-ADTContentToCache

DanGough:
# Pull Request

## Description

- Currently Copy-ADTContentToCache copies to a cache location that defaults to C:\ProgramData\SoftwareCache.
- A user could pre-create this cache folder, or the package-specific subfolder, which would give them write access.
- Package-specific folders are not erased before copying, allowing for malicious content to be placed here before copying takes place.
- A non-admin user could therefore poison the cache used by system context installs or other users.
- Running Copy-ADTContentToCache in user context can create a cached folder that is not writable by other users, leading to a failure when another user tries to copy content to the same location.

To resolve:

- A new config element has been provided to set a separate cache location for use when the caller does not have admin rights (mirroring the setup we have currently for log folders etc).
- Destination is erased prior to copying (this is potentially risky if a user does something stupid like override the LiteralPath with C:\Program Files... but I'm not sure how to otherwise improve on this other than having a list of disallowed locations?).
- To prevent the case where the all-user cache was pre-created by a user, we set the owner to the Administrators group and reset ACLs (achieved by adding an `-Owner` parameter to `Set-ADTItemPermission` - I did at first consider a new `Set-ADTItemOwner` function but decided to roll it into the existing one instead)

In addition, this reworks the `-Exclude` parameter to use `-Content`, so that we are free to add Include/Exclude parameters in future that use actual filters like `Copy-Item`, a feature that was recently requested in https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/2265.

## Type of change

## Checklist

## How Has This Been Tested?

- Verified that non-admins end up writing to a cache under their own user profile instead of the shared one.
- Verified that existing content gets erased before copying.
- Verified that if a non-admin user pre-creates the admin cache location under ProgramData, running the command afterwards as admin hardens the cache location and the user can no longer write there.


src/PSADT/PSADT/FileSystem/FileSystemUtilities.cs
```diff
@@ -91,6 +92,49 @@ public static void SetAccessControl(DirectoryInfo directoryInfo, DirectorySecuri
...
+        /// <remarks>The <see cref="SE_PRIVILEGE.SeTakeOwnershipPrivilege"/> is required to reassign
+        /// ownership when the caller does not already own the item (for example, when a folder was pre-created by
+        /// another account). This privilege is held by the SYSTEM account and members of the local Administrators
+        /// group.</remarks>
+        /// <param name="info">The file or directory whose owner is to be set.</param>
+        /// <param name="identity">The identity to set as the owner of the item.</param>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="info"/> or <paramref name="identity"/> is <see langword="null"/>.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the current process does not hold the required <see cref="SE_PRIVILEGE.SeTakeOwnershipPrivilege"/>.</exception>
+        /// <exception cref="ArgumentException">Thrown when <paramref name="info"/> is not a supported file system object.</exception>
+        public static void SetOwner(FileSystemInfo info, IdentityReference identity)
```
mjr4077au: I'd probably prefer separate `DirectoryInfo` and `FileInfo` overloads like everything else in `FileSystemUtilities`. What's here works, its just not consistent with everything else.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -123,14 +124,84 @@ function Copy-ADTContentToCache
...
+        {
+            try
+            {
+                try
+                {
+                    Write-ADTLogEntry -Message "Securing root cache folder [$cachePath]."
+                    if ($PSCmdlet.ShouldProcess($cachePath, 'Secure root cache folder'))
+                    {
+                        # Using the SID instead of BUILTIN\Administrators to overcome localization issues.
+                        Set-ADTItemPermission -LiteralPath $cachePath -Owner '*S-1-5-32-544' -EnableInheritance -RemoveExplicitRules -InformationAction SilentlyContinue
```
mjr4077au: I know it's more setup, but using the value from `[System.Security.Principal.SecurityIdentifier]::new([System.Security.Principal.WellKnownSidType]::BuiltinAdministratorsSid, $null).Value` is probably more ideal than a magic constant as it's self-documenting.

src/PSAppDeployToolkit/Public/Copy-ADTContentToCache.ps1
```diff
@@ -39,17 +41,17 @@ function Copy-ADTContentToCache
...
 
     .EXAMPLE
-        Copy-ADTContentToCache -Exclude Files,SupportFiles
+        Copy-ADTContentToCache -Content Toolkit
 
         This example copies the toolkit content to the default cache folder, excluding the Files and SupportFiles folders and leaving DirFiles and DirSupportFiles pointing at the original location.
 
     .EXAMPLE
-        Copy-ADTContentToCache -Exclude Toolkit
+        Copy-ADTContentToCache -Content Files,SupportFiles
```
mjr4077au: I'd have a space after the comma for correct punctuation, etc.

mjr4077au (changes requested):
I only have a few nits, otherwise looks good. I've resolved all Copilot feedback that was incorrect, but there's a few of its findings which are valid and should be addressed.

DanGough:
All recommendations implemented, also tweaked parameter sets so that Set-ADTItemPermission can change the owner only without any other ACL changes being involved.

mjr4077au (approved):
Nice one. LGTM!

PR #2274, opened 2026-08-07
Add identity transformer

neo42JBR:
# Pull Request

## Description

Added a identity transformation attribute for consumtion in functions that require identity information.
It can handle transforming inputs from common sources  into a IdentityReference for consumption within a function.

**Sample**
```powershell
function test {
  param(
    [PSAppDeployToolkit.Attributes.IdentityTransformation()]
    [System.Security.Principal.IdentityReference]
    $User
  )
  return $User.Translate([System.Security.Principal.NTAccount]).Value
}
```

This will allow inputs like:
```powershell
test -User (Get-LocalUser -Name 'Administrator')
test -User ([System.Security.Principal.WindowsIdentity]::GetCurrent())
test -User ([System.Security.Principal.SecurityIdentifier]::new('S-1-5-18'))
test -User ([System.Security.Principal.WellKnownSidType]::LocalSystemSid)
test -User 'S-1-5-18'
test -User 'SYSTEM'
test -User 'LocalSystemSid'
```

Within code, can be translated into any `IdentityReference` with the `Translate()` method.

This should streamline handling identities.

## Type of change

## Checklist

- [ ] I have tested my changes to prove my fix is effective

## How Has This Been Tested?

This class has been tested in downstream projects from my company. It was adapted to fit the coding style of the PSADT

### DIFF
--- a/src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
+++ b/src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using PSAppDeployToolkit.Utilities;
+using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Management.Automation;
 using System.Reflection;
@@ -27,31 +28,31 @@ namespace PSAppDeployToolkit.Attributes
         /// <exception cref="ArgumentException">Thrown if the input object is not supported for transformation.</exception>
         /// <exception cref="ArgumentNullException">Thrown if the input data is null.</exception>
         [SuppressMessage("Style", "IDE0046:Use conditional expression for return", Justification = "Conditional expression would be unreadable.")]
-        public override object Transform(EngineIntrinsics engineIntrinsics, object inputData)
+        public override object Transform(EngineIntrinsics engineIntrinsics, object? inputData)
         {
-            if (inputData is null)
+            if (!PowerShellUtilities.TryGetBaseObject(inputData, out inputData))
             {
-                throw new ArgumentNullException(nameof(inputData), "Cannot transform null to an IdentityReference.");
-            }
-            if (inputData is WellKnownSidType wellKnownSidType)
-            {
-                return new SecurityIdentifier(wellKnownSidType, domainSid: null);
+                throw new ArgumentNullException(paramName: nameof(inputData), "Cannot transform null to IdentityReference.");
             }
             if (inputData is IdentityReference identity)
             {
                 return identity;
             }
+            if (inputData is WellKnownSidType wellKnownSidType)
+            {
+                return new SecurityIdentifier(wellKnownSidType, domainSid: null);
+            }
             if (inputData is WindowsIdentity windowsIdentity)
             {
                 return windowsIdentity.User ?? throw new InvalidOperationException("The given WindowsIdentity did not provide the required user identity.");
             }
-            if (TryConvertPowerShellCommandObject(inputData, out SecurityIdentifier? pwshCommandSid))
-            {
-                return pwshCommandSid;
-            }
             if (inputData is string str && TryParseIdentityReference(str, out IdentityReference? strIdentity))
             {
                 return strIdentity;
+            }
+            if (TryConvertPowerShellCommandObject(inputData, out SecurityIdentifier? pwshCommandSid))
+            {
+                return pwshCommandSid;
             }
             throw new ArgumentException("Input data must be of type IdentityReference, WindowsIdentity, LocalPrincipal, WellKnownSidType or a string representing any of those types.", nameof(inputData));
         }
@@ -96,10 +97,6 @@ namespace PSAppDeployToolkit.Attributes
         private static bool TryParseSid(string identityString, [NotNullWhen(true)] out SecurityIdentifier? identity)
         {
             identity = null;
-            if (string.IsNullOrWhiteSpace(identityString))
-            {
-                return false;
-            }
             try
             {
                 identity = new SecurityIdentifier(identityString);
@@ -120,10 +117,6 @@ namespace PSAppDeployToolkit.Attributes
         private static bool TryParseNTAccount(string identityString, [NotNullWhen(true)] out NTAccount? identity)
         {
             identity = null;
-            if (string.IsNullOrWhiteSpace(identityString))
-            {
-                return false;
-            }
             try
             {
                 identity = new NTAccount(identityString);
@@ -145,15 +138,6 @@ namespace PSAppDeployToolkit.Attributes
         private static bool TryConvertPowerShellCommandObject(object identityObject, [NotNullWhen(true)] out SecurityIdentifier? identity)
         {
             identity = null;
-            // Unwrap the actual object, if it is wrapped
-            if (identityObject is PSObject psObj)
-            {
-                identityObject = psObj.BaseObject;
-            }
-            if (identityObject is null)
-            {
-                return false;
-            }
             Type objectType = identityObject.GetType();
             if (!string.Equals(objectType.Namespace, "Microsoft.PowerShell.Commands", StringComparison.Ordinal))
             {
@@ -166,12 +150,12 @@ namespace PSAppDeployToolkit.Attributes
                 identity = directSid;
                 return true;
             }
-            if (objectType.BaseType is Type parentType
-                && string.Equals(parentType.Name, "LocalPrincipal", StringComparison.Ordinal)
-                && objectType.GetProperty("SID", typeof(SecurityIdentifier)) is PropertyInfo parentProperty
-                && parentProperty.GetValue(identityObject) is SecurityIdentifier parentSid)
+            if (objectType.BaseType is Type baseType
+                && string.Equals(baseType.Name, "LocalPrincipal", StringComparison.Ordinal)
+                && objectType.GetProperty("SID", typeof(SecurityIdentifier)) is PropertyInfo baseProperty
+                && baseProperty.GetValue(identityObject) is SecurityIdentifier baseSid)
             {
-                identity = parentSid;
+                identity = baseSid;
                 return true;
             }
             return false;
--- a/src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
+++ b/src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Management.Automation;
 using System.Reflection;
 using System.Security.Principal;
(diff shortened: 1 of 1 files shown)


src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+        ///  * <see cref="string"/> representing the above
+        /// </summary>
+        /// <param name="engineIntrinsics">The PowerShell engine intrinsics.</param>
+        /// <param name="inputData">The input value to transform.</param>
+        /// <returns></returns>
+        /// <exception cref="InvalidOperationException">Thrown if the input object did not contain the required properties.</exception>
+        /// <exception cref="ArgumentException">Thrown if the input object is not supported for transformation.</exception>
+        public override object Transform(EngineIntrinsics engineIntrinsics, object inputData)
+        {
+
```
mjr4077au: Empty white line here.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                return new SecurityIdentifier(wellKnownSidType, null);
+            }
+            else if (inputData is IdentityReference identity)
+            {
+                return identity;
+            }
+            else if (inputData is WindowsIdentity windowsIdentity)
+            {
+                return windowsIdentity.User
```
mjr4077au: Not keen on broken lines like this. We're all on wide monitors these days.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+                    ?? throw new InvalidOperationException("The given WindowsIdentity did not provide the required user identity.");
+            }
+            else if (inputData.GetType().FullName == "Microsoft.PowerShell.Commands.LocalUser"
+                && inputData.GetType().GetProperty("SID", typeof(SecurityIdentifier)) is PropertyInfo localUserSidProperty)
```
mjr4077au: Same here.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                return localUserSidProperty.GetValue(inputData)
+                    ?? throw new InvalidOperationException("The given LocalUser did not provide the required SID property.");
```
mjr4077au: And here.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Tries to parse a string as <see cref="IdentityReference"/>, returning the corresponding object.
+        /// </summary>
+        /// <param name="identityString">The string to parse.</param>
+        /// <param name="identity">The <see cref="IdentityReference"/> if parsing succeeds; otherwise NullSid</param>
+        /// <returns>True if the input string was successfully parsed  name; otherwise, false.</returns>
+        private static bool TryParseIdentityReference(string identityString, out IdentityReference identity)
```
mjr4077au: This should have a `NotNullWhen(true)]` decoration on the out'd value.

mjr4077au: This should be `out IdentityReference?`

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            identity = new SecurityIdentifier(WellKnownSidType.NullSid, null);
```
mjr4077au: Don't default initialise like this, set `identity` as null when we're returning false.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+            if (identityString.StartsWith("S-", StringComparison.OrdinalIgnoreCase))
+            {
+                try
+                {
+                    identity = new SecurityIdentifier(identityString);
+                    return true;
+                }
+                catch (ArgumentException)
+                {
+                }
```
mjr4077au: Don't use empty catch blocks like this, explicitly return false from here.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            // Try NTAccount
+            try
+            {
+                identity = new NTAccount(identityString);
+                return true;
+            }
+            catch (IdentityNotMappedException)
+            {
```
mjr4077au, line 126: Same here also.

mjr4077au (changes requested):
I've left some review comments for your consideration. Has this been tested to see whether it compiles locally? There's a lot of things here our analysers should trip up on.

neo42JBR:
I reworked the attribute, including your change requests.
The logic has been split into multiple methods in order to accomadate these changes.

In addition I updated supported from LocalUser to LocalPrincipal (The parent type).

I built and tested locally.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// Try to extract the SID from a LocalPrincipal type using reflection. The LocalPrincipal type is used within the Microsoft.PowerShell.LocalAccounts module.
+        /// </summary>
+        /// <param name="identityObject">The object to analyse.</param>
+        /// <param name="identity">The parsed SecrutiyIdentifier.</param>
+        /// <returns>True if extraction was successful, otherwise false.</returns>
+        private static bool TryConvertPowerShellCommandObject(object identityObject, [NotNullWhen(true)] out SecurityIdentifier? identity)
+        {
+            identity = null;
+            // Unwrap the actual object, if it is wrapped
+            if (identityObject is PSObject psObj)
```
mjr4077au: Our other extensions use the following to unwrap the object:
```c#
if (!PowerShellUtilities.TryGetBaseObject(inputData, out inputData))
{
    throw new ArgumentNullException(paramName: null, "Cannot transform null to DateTime.");
}
```
This is preferred as it recursively unwraps PSObject types that might be wrapping another PSObject (rare, but possible).

This should also be done within `public override object Transform(EngineIntrinsics engineIntrinsics, object inputData)`, not at this late stage.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <param name="engineIntrinsics">The PowerShell engine intrinsics.</param>
+        /// <param name="inputData">The input value to transform.</param>
+        /// <returns>The identity reference the input object represents.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if the input object was valid but did not contain the required data.</exception>
+        /// <exception cref="ArgumentException">Thrown if the input object is not supported for transformation.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if the input data is null.</exception>
+        [SuppressMessage("Style", "IDE0046:Use conditional expression for return", Justification = "Conditional expression would be unreadable.")]
+        public override object Transform(EngineIntrinsics engineIntrinsics, object inputData)
+        {
+            if (inputData is null)
```
mjr4077au: As mentioned within `private static bool TryConvertPowerShellCommandObject(object identityObject, [NotNullWhen(true)] out SecurityIdentifier? identity)`, the PSObject unwrapping should occur here and be done with our utilities we already have. This will test nullability also.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+            if (inputData is IdentityReference identity)
+            {
+                return identity;
+            }
+            if (inputData is WindowsIdentity windowsIdentity)
+            {
+                return windowsIdentity.User ?? throw new InvalidOperationException("The given WindowsIdentity did not provide the required user identity.");
+            }
+            if (TryConvertPowerShellCommandObject(inputData, out SecurityIdentifier? pwshCommandSid))
```
mjr4077au: I'd prefer the reflection pathway be a last-ditch effort, so please make this branch occur last.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+        {
+            if (inputData is null)
+            {
+                throw new ArgumentNullException(nameof(inputData), "Cannot transform null to an IdentityReference.");
+            }
+            if (inputData is WellKnownSidType wellKnownSidType)
+            {
+                return new SecurityIdentifier(wellKnownSidType, domainSid: null);
+            }
+            if (inputData is IdentityReference identity)
```
mjr4077au, line 37: This should be considered the happy path, so move it above `if (inputData is WellKnownSidType wellKnownSidType)`.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <returns>True if extraction was successful, otherwise false.</returns>
+        private static bool TryConvertPowerShellCommandObject(object identityObject, [NotNullWhen(true)] out SecurityIdentifier? identity)
+        {
+            identity = null;
+            // Unwrap the actual object, if it is wrapped
+            if (identityObject is PSObject psObj)
+            {
+                identityObject = psObj.BaseObject;
+            }
+            if (identityObject is null)
```
mjr4077au: With the PSObject unwrapping, this null check won't be needed as the input will be tested and confirmed non-null before proceeding.

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <summary>
+        /// Try to parse string into a <see cref="SecurityIdentifier"/>.
+        /// </summary>
+        /// <param name="identityString">The string representing the SecrutiyIdentifier.</param>
+        /// <param name="identity">The parsed SecrutiyIdentifier.</param>
+        /// <returns>True if parsing was successful, otherwise false.</returns>
+        private static bool TryParseSid(string identityString, [NotNullWhen(true)] out SecurityIdentifier? identity)
+        {
+            identity = null;
+            if (string.IsNullOrWhiteSpace(identityString))
```
mjr4077au: We're doing this null or whitespace test here but it's already done by the caller (`TryParseIdentityReference()`).

src/PSADT/PSAppDeployToolkit/Attributes/IdentityTransformationAttribute.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <summary>
+        /// Try to parse string into a <see cref="NTAccount"/>.
+        /// </summary>
+        /// <param name="identityString">The string representing the NTAccount.</param>
+        /// <param name="identity">The parsed NTAccount.</param>
+        /// <returns>True if parsing was successful, otherwise false.</returns>
+        private static bool TryParseNTAccount(string identityString, [NotNullWhen(true)] out NTAccount? identity)
+        {
+            identity = null;
+            if (string.IsNullOrWhiteSpace(identityString))
```
mjr4077au: We're doing this null or whitespace test here but it's already done by the caller (`TryParseIdentityReference()`).

mjr4077au (changes requested):
Looking a lot better, at least it's compiling now! Just left a few more comments, mostly minor stuff but would appreciate you taking a look.

Once done, this should be good to merge and would be great to use within functions like `Set-ADTItemPermission`, etc.

neo42JBR:
I applied the requested changes.

One use-case could also be within "ConvertTo-ADTNTAccountOrSID"
Merge AccountName, SID, WellKnownSidName into "InputObject" (keep alias for backwards compatibility?)

mjr4077au (approved):
Thanks for all of that, looks great now. And I agree with the suggested usages and would welcome a PR if you're keen.

It's not important for this PR, but if we do hook it up within `Set-ADTItemPermission`, we'll need to expand `TryParseSid()` to trim any leading asterisks since that's what that function uses to denote a SID (I believe it's an `icacls`-ism).

PR #2280, opened 2026-08-13
Add `-Force` to `Get-ADTApplication` and `Uninstall-ADTApplication`.

mjr4077au:
# Pull Request

## Description

* Add new parameters to `Get-ADTApplication` function.
* Remove `InstalledApplication.Is64BitApplication` property.
* Update all NuGet packages and rebuild debug DLLs.

The `Is64BitApplication` one probably needs some discussion. I've been doing a lot of apps stuff at work and such a value based purely on the WOW6432Node location of the uninstall entry is not a strong indicator of the installed application's bitness.

* For MSIs, as msiexec.exe installs it, it'll always have its uninstall entry reside in the system's native/64-bit registry location.
* For Burn installers, their installer is always 32-bit so will be in the WOW6432Node area even if the app or MSI is 64-bit.

This property and its name/description is asserting with practically complete confidence as to the architecture of the installed application purely by where the uninstall entry resides in the registry, which isn't right and isn't accurate.

## Type of change

## Checklist

## How Has This Been Tested?

```
PS C:\Repos\PSAppDeployToolkit> Import-Module -Name C:\Repos\PSAppDeployToolkit\src\PSAppDeployToolkit
PS C:\Repos\PSAppDeployToolkit> $apps = Get-ADTApplication
PS C:\Repos\PSAppDeployToolkit> $apps.Count
42
PS C:\Repos\PSAppDeployToolkit> $apps = Get-ADTApplication -IncludeNoRemoveEntries
PS C:\Repos\PSAppDeployToolkit> $apps.Count
43
PS C:\Repos\PSAppDeployToolkit> $apps = Get-ADTApplication -IncludeSystemComponentEntries
PS C:\Repos\PSAppDeployToolkit> $apps.Count
246
PS C:\Repos\PSAppDeployToolkit>
```


neo42JBR:
Just leaving my two cents here:

* Regarding `-IncludeNoRemoveEntries` and `-IncludeSystemComponentEntries`
  I like the feature, but would recommend using a negated approach (include by default) to not introduce a breaking change.

* Regarding `Is64BitApplication`
  I am opposed to the removal. It is the only good point of data to know the registry origin. 
  You are correct that it does not reflect the application architecture, but it is vital to be able differentiate apps that do use both keys

mjr4077au:
@neo42JBR there's been changes since your last comment, such as the `Is64BitApplication` reversion. Please feel free to review and reply, but we do need to make sure the command is safe in its default state. Having `Get-ADTApplication` include `SystemComponent` apps by default was a design mistake that shouldn't be left as it was.

DanGough (approved):

DanGough:
@neo42JBR behaviour is now:

Get-ADTApplication requires -Force to return hidden SystemComponent entries (much like Get-Item requires -Force to return hidden files)

Uninstall-ADTApplication requires -Force to uninstall hidden SystemComponent entries, or entries marked as NoRemove (like Remove-Item requires -Force to delete hidden or read-only files)

I can appreciate this may be a breaking change in some rare cases, but all breaking changes will be fully documented. It does provide benefits, e.g. ensuring that by default uninstall would target a WiX burn wrapper EXE entry rather than the hidden MSI entry that it installed.

DanGough:
FYI - Issue found that was resolved in a subsequentr PR - apps not being returned due to `The variable '$noRemove' cannot be retrieved because it has not been set.`: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/pull/2281

Issue #2301, opened 2026-08-20
[Bug] Start-ADTMsiProcess add user to logname even elevated


AlkHacNar:
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.2.0 RC1

### Describe the bug

The msi logs append the username to the log in an elevated session, adtSession even have requireAdmin = $true set. I use  -LogFileName $adtSession.msiLogName, but even withput it, it still appends it.

#1979 was the exact opposite and maybe there was a problem?

<img width="1107" height="381" alt="Image" src="https://github.com/user-attachments/assets/835ffb17-2f32-48b6-9467-062279909e46" />

<img width="2440" height="272" alt="Image" src="https://github.com/user-attachments/assets/8e9781c9-ce35-4f59-9309-d7d296a90654" />

### Steps to reproduce

1.open Terminal as Admin and start or start Invoke-AppDeployToolkit.exe as admin directly
2.MSI Log have ther appended to logfilename

### Environment data

```console
OsName               : Microsoft Windows 11 Enterprise
OSDisplayVersion     : 25H2
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-Bit
WindowsVersion       : 2009
WindowsProductName   : Windows 10 Enterprise
WindowsBuildLabEx    : 26100.1.amd64fre.ge_release.240331-1435
OsLanguage           : de-DE
OsMuiLanguages       : {de-DE, en-US}
KeyboardLayout       : de-DE
TimeZone             : (UTC+01:00) Amsterdam, Berlin, Bern, Rom, Stockholm, Wien
HyperVisorPresent    : True
CsPartOfDomain       : False
CsPCSystemType       : Mobile


Host:
  Version:      8.0.28
  Architecture: x64
  Commit:       46295af582
  RID:          win-x64

.NET SDKs installed:
  No SDKs were found.

.NET runtimes installed:
  Microsoft.NETCore.App 8.0.28 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
[...]

PR #2318, opened 2026-08-29
Implement long-awaited unit testing for entire C# estate.

### LINKED ISSUE #1414: [Feature] Write unit/integration tests for everything to improve toolkit reliability

### LINKED ISSUE #2328: [Bug] Test-ADTUserInFocusMode crash on Windows Server 2022
### Prerequisites

- [x] Ensure you write a short, descriptive title after [Bug] above.
- [x] Make sure to [search for any existing issues](https://github.com/psappdeploytoolkit/psappdeploytoolkit/issues) before filing a new one.
- [x] Verify you are able to reproduce the issue with the [latest released version](https://www.github.com/psappdeploytoolkit/psappdeploytoolkit/releases/latest)

### PSAppDeployToolkit version

4.2.0-RC1

### Describe the bug

`PSADT.WindowsRuntime.UI.Shell.ShellUtilities.TryGetFocusSessionActive` throws `System.ApplicationException` on Windows Server 2022.

The type `Windows.UI.Shell.FocusSessionManager` does not exist. 

### Steps to reproduce

1. Run `Test-ADTUserInFocusMode` on Windows Server 2022

### Environment data

```console
OsName               : Microsoft Windows Server 2022 Standard
OSDisplayVersion     : 21H2
OsOperatingSystemSKU : StandardServerEdition
OsArchitecture       : 64-Bit
WindowsVersion       : 2009
WindowsProductName   : Windows Server 2022 Standard
WindowsBuildLabEx    : 20348.1.amd64fre.fe_release.210507-1500
OsLanguage           : de-DE
OsMuiLanguages       : {de-DE}
KeyboardLayout       : de-DE
TimeZone             : (UTC+01:00) Amsterdam, Berlin, Bern, Rom, Stockholm, Wien
HyperVisorPresent    : True
CsPartOfDomain       : True
CsPCSystemType       : Mobile

.NET SDK:
 Version:           10.0.400-preview.0.26322.102
 Commit:            0cf89a94a1
 Workload version:  10.0.400-manifests.11c77abd
 MSBuild version:   18.9.0-preview-26322-102+0cf89a94a

Laufzeitumgebung:
 OS Name:     Windows
 OS Version:  10.0.20348
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\10.0.400-preview.0.26322.102\

Installierte .NET-Workloads:
Es sind keine installierten Workloads zum Anzeigen vorhanden.
Für die Verwendung von workload sets bei der Installation neuer Manifeste konfiguriert.
Es sind keine Workloadsätze installiert. Führen Sie „dotnet workload restore“ aus, um einen Workloadsatz zu installieren.

Host:
  Version:      10.0.9
  Architecture: x64
  Commit:       901ca94124

.NET SDKs installed:
  8.0.303 [C:\Program Files\dotnet\sdk]
  8.0.423 [C:\Program Files\dotnet\sdk]
  10.0.201 [C:\Program Files\dotnet\sdk]
  10.0.400-preview.0.26322.102 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.32 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.28 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.29 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 10.0.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 10.0.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.32 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.28 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.29 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 10.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 10.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 6.0.32 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.7 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.28 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.29 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 10.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 10.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  x86   [C:\Program Files (x86)\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  Not set

global.json file:
  Not found

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
  https://aka.ms/dotnet/download
```

mjr4077au:
# Pull Request

## Description

This PR brings long-awaited unit testing to our entire C# estate, including over 6,500 unit tests (~3,200 per TFM) across all projects, bringing us up to an average code coverage percentage value of 74.4%.

image

| Assembly | Covered | Lines | Coverage |
| --- | ---: | ---: | ---: |
| `PSADT.UserInterface` | 749 | 772 | 97.0% |
| `PSADT.ClientServer.Server` | 866 | 988 | 87.7% |
| `PSAppDeployToolkit` | 1,596 | 1,960 | 81.4% |
| `PSADT.CodeAnalysis` | 24 | 30 | 80.0% |
| `PSADT` | 5,408 | 7,116 | 76.0% |
| `PSADT.UserInterface.Interfaces` | 2,364 | 3,216 | 73.5% |
| `PSADT.WindowsRuntime` | 10 | 16 | 62.5% |
| `PSADT.ClientServer.Client.Compatible` | 279 | 636 | 43.9% |
| `PSADT.ClientServer.Client` | 141 | 636 | 22.2% |
| **Total** | **11,437** | **15,370** | **74.4%** |

As part of this, the repository has undergone a slight re-arrangement into a more dotnet-compliant structure as documented here: https://learn.microsoft.com/en-us/dotnet/core/porting/project-structure. This allows us to have test-specific `Directory.Build.props` setup which greatly simplifies repository management.

The new layout is also now represented in the .slnx file and visually within Visual Studio:

image

Also included is uplifted Pester tests, covering areas not covered before, as well as expanding on some other unit tests. Crucial elements like `Start-ADTProcess` are now stressed, and `Start-ADTMsiProcess` is also courtesy of a nice little test MSI from Master Packager. Code Coverage on the PowerShell side went from ~20% to ~67%. The remainder are system-changing functions or UI elements that cannot be easily tested.

Over 90 defects were uncovered and remediated as part of this cycle, a worthy grind to complete before the final 4.2.0 release to put our best foot forward.

Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/1414.
Fixes https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/issues/2328.

### Overall changes:

- Implement unit testing for entire C# project state within this repository.
- Slightly re-arrange repository to get it into a more dotnet-compliant state.
- Update module build system to account for changes in newer dotnet SDK versions.
- Extract duplicated xunit.v3 .csproj settings into their own Directory.Build.props file.
- Re-organise .slnx files so that the layout in the tree represents repo structure.
- Update our .gitignore file to the latest template provided to us by GitHub.
- Remove any unused Meziantou.Polyfill polyfills that were being generated.
- Rework existing `PSADT.Tests` so that they target both our supported TFMs.
- Don't let our project be build using unsupported dotnet SDK versions.
- Add `PointerTypedConstant<TSelf>` to reduce `TypedConstant<TSelf>` scope.
- Add unused, but missing `LSA_POLICY_ACCESS.POLICY_NOTIFICATION` enum member.
- Add some Pester testing for our C#-based ValidationAttribute types.
- Make `LogUtilities.LogFileNameRegex` regex case-insensitive.
- Rename `RestartDialogOptions.AllowCancel` to `DialogAllowCancel.
- Collapse all `PSADT.ClientServer.Client` projects into one with some magic.
- Reuse `ClientExecutable`'s wrapper deserialisation methods everywhere.
- Remove unnecessary MA0074 suppression for all projects.
- Update Fluence.Wpf to the latest version, mostly for NuGet update.
- Update all NuGet packages to their latest versions, addressing all warnings.

### Fixed issues:

- Add missing overflow checks to all `TypedConstant<TSelf>` converter methods.
- Fix `NullableExtensions.ToPointer<T>()` that was taking input by value.
- Fix range test in `ArgumentOutOfRangeExceptionExtensions.ThrowIfZeroOrInvalid()`.
- Correct a bunch of range checks across `SafeMemoryHandle<TSelf>` implementation.
- Fix bad setup within `ShortcutHotkey`, utility class used by `ShellLinkFile`.
- Fix athematic overflow within `NativeUtilities.Load64()` (only impacts ARM64).
- Rework a number of `record` objects that did not have true IEquality status.
- Add input validation to a number of public and internal methods/ctors.
- Correct the text on some thrown exceptions within the PSADT project.
- Amend `IPersistFileExtensions.GetCurFile()` to test for successful failure.
- Fix some calls to `NativeMethods.MsiViewFetch()` without `ERROR_NO_MORE_ITEMS` tests.
- Rework awkward `BatteryInfo` class setup into a proper singleton object.
- Make `WindowUtilities.GetProcessWindowInfo()` gracefully skip closed windows.
- Amend `EnvironmentUtilities.GetEnvironmentVariables() due to net472/net8.0 differences.
- Amend `ProcessManager.LaunchAsync()` stdin writing to ensure no BOM is ever written.
- Make sure `RebootInfo` factors in `IsIntuneClientRebootPending` as it was missed.
- Fix inconsistent return value from `InternetShortcutFile.IconFile` property.
- Make `AppxManifestInfo.ResourceId` string property nullable as its not guaranteed.
- Fix bad flags being passed on within `MsiUtilities.CreatePropertyTransformationFile()`.
- Delete broken/unused `DateTimeTransformationAttribute` before it makes it to release.
- Ensure all PSADT.ClientServer.Server payloads are referenced in `DataSerialization`.
- Fix bad Dictionary count check in `BaseValidateNotEmptyOrWhiteSpaceAttribute`.
- Fix missing `else if` usage within `DeploymentSession` class constructor.
- Correct string value parsing within `TimeSpanTransformationAttribute` class.
- Add null check in `PowerShellUtilities.TryGetBaseObject<T>()` helper method.
- Fix bad Dictionary value check in `BaseValidateNotEmptyOrWhiteSpaceAttribute`.
- Make `ValidateUniqueAttribute` not be dependent on processing order.
- Fix bad enum parsing in `IdentityReferenceTransformationAttribute`.
- Don't use `ValidateNotNullOrWhiteSpaceAttribute` on enum types.
- Improve `BaseValidateNotEmptyOrWhiteSpaceAttribute` empty checks.
- Fix bad double space replacement setup in `DeploymentSession`.
- Harden `DeploymentSession` log name formatting from bad input.
- Apply same log name format hardening within session closure also.
- Improve `ServerInstance` class disposal if issue occurs within finally block.
- Fix double serialisation issue for `CustomDialogResult.Result` property.
- Remove the `[Flags]` decoration from three dialog enums that aren't bitfields.
- Add missing percentage range checks for `Show-ADTInstallationProgress`.
- Fix `ListSelectionBox.DialogTopMost` boolean inconsistency with others.
- Properly verify `RestartDialogOptions.ShutdownReasonText` isn't whitespace.
- Remove dead `InternalsVisibleTo` to non-existent/renamed C# project.
- Fix bad/missing classic form subtypes in `PSADT.UserInterface.Interfaces.csproj`.
- Remove redundant divide and multiply math from `ClassicDialog.PositionForm()`.
- Fix bad parameter nullability for `IntToVisibilityConverter` WPF utility class.
- Ensure `ProgressDialog.UpdateProgress()` doesn't reset text alignment.
- Fix bad checking order within `DialogManager.UpdateProgressDialogAsync()`.
- Apply same checking fix for `DialogManager.UpdateNotifyIconAsync()` method.
- Simplify `ClassicDialog.MinimizeBox` setting and remove redundancy.
- Add missing `SelectedIndex = -1` within `ListSelectionBox` class.
- Ensure `ClientExecutable.ArgvToDictionary()` validates param names.
- Ensure `ClientExecutable.ArgvToDictionary()` doesn't allow duplicates.
- Strengthen `ClientExecutable.ArgvToDictionary()` registry operations.
- Add `string[]` to fix `System.Xml.XmlException` serialisation.
- Fix potential infinite looping in `Test-ADTServiceExists.Tests.ps1`.
- Fix bad initialisation within `ExportedFunctions.Tests.ps1`.
- Fix two Pester tests to handle Windows PowerShell/PowerShell 7 differences.
- Skip the font tests when unelevated instead of failing on `#Requires`.
- Treat a null remaining-arguments list as no arguments.
- Sweep orphaned test fonts either side of the font tests.
- Recurse into the function rather than the scriptblock it is written in.
- Describe a rejected value only when there is something to describe.
- Guard the export against any existing path, not just a directory.
- Take an MSI exit code as unsigned, matching WIN32_ERROR.
- Iterate the group policy accounts when resolving a name to a SID.
- Skip a profile whose SID has no account rather than erroring on it.
- Confirm a file is a PE image before reading its machine type.
- Return the deferral run interval alongside the rest of the history.
- Set the culture on the thread rather than only in PowerShell's cache.
- Check for subkeys under the keys a wildcard resolved to.
- Call an overload that exists when no environment target is given.
- Apply the append and remove options to the process scope as well.
- Keep the module directories PowerShell added for itself when refreshing.
- Only pass a transform to build on when the caller named one.
- Seed the Edge extension policy with a value that can be read back.
- Apply an access control list that was read from somewhere else.
- Bind MSI property values rather than writing them into the statement.
- Name the shortcut tests after the functions they test.
- Give two tests something they can actually fail against.
- Skip the service dependency tests where there is nothing to try them on.
- Cover the paths the existing tests left out.
- Withhold a secured argument list from the bound parameter dump.
- Escape a quote inside a session property value that stays live.
- Refuse a service name pattern only in CIM mode.
- Drop the branches for clearing a shortcut's target.
- Clear a shell link's icon with an empty location, not none.
- Name whoami by its path when checking the caller's token.
- Drop an internet shortcut's working directory and window style.
- Take a process id as unsigned, and name a Win32 error the same way twice.
- Stop a switch being passed to two commands that never had it.
- Drop a guard against a parameter the function does not have.
- Resolve an uninstall program named without a path against System32.
- Make `ShellLinkFile.Resolve()` method work as it's meant to.

## Type of change

## Checklist

- [ ] I am pulling to the **main** branch
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have tested my changes to prove my fix is effective
- [ ] I have tested that the module can build following my changes
- [ ] I have made sure that any script file-encoding is set to UTF8 with BOM, i.e. unchanged.

## How Has This Been Tested?

Full module build, including existing Pester tests. Dry run through a number of scenarios.


DanGough (commented):
Found an issue pre-existing in RC1, found here when testing changes to SecureArgumentList handling in header functions (error only shows with below command when used with `-SecureArgumentList`):

```powershell
Start-ADTProcess -FilePath C:\Windows\System32\cmd.exe -ArgumentList '/c exit 1' -SecureArgumentList
```
```
Invoke-ADTFunctionErrorHandler : Parameter set cannot be resolved using the specified named parameters.
At C:\Users\admin\Desktop\PSAppDeployToolkit\Public\Start-ADTProcess.ps1:1098 char:17
+                 Invoke-ADTFunctionErrorHandler @iafehParams -Silent
```

Another thing (also found in AI report below) was that Uninstall-ADTApplication now only looks under System32 if the uninstaller path is not fully qualified, when before it searched the whole system path. I can put `notepad` as my uninstall string and add/remove programs will execute it, as did PSADT 4.1; now it complains `Could not find file 'C:\WINDOWS\system32\notepad'`.


Opus 5 review of all changes after 8cc451353c075156e8455ee53665ed9c491e5962 is below. Most of the 'Breaking changes' are intentional and can be ignored, but I posted the entire report:

## Likely bugs (highest confidence)

**1. `-Append`/`-Remove` read from the wrong scope — data-loss risk on machine PATH** — `ceefbad6e`
`EnvironmentUtilities.cs:186` and #L211 compute the append/remove result from `GetEnvironmentVariable(variable)` — the **process** environment — then write it to the `target` registry hive. So `Set-ADTEnvironmentVariable -Variable PATH -Value X -Append -Target Machine` writes the *process's* fully-expanded PATH (which includes the user hive and anything the session added) into `HKLM`. The read was already target-less before this commit, but the commit's stated intent ("append/remove now apply to a process variable the same way as user or machine") makes the mismatch worse, not better. Should be `GetEnvironmentVariable(variable, target)`.

**2. `ShellLinkInfo.IconLocation` mangles environment-variable icon paths** — `448f658b9`
`IconLocationValue = shellLink.IconLocation?.FullName`. `IShellLinkW::GetIconLocation` returns the raw string, which is very commonly `%SystemRoot%\System32\imageres.dll`. Taking `.FullName` prefixes the current working directory, so `Get-ADTShortcut` now reports `D:\SomeCwd\%SystemRoot%\System32\imageres.dll`. Previously the `FileInfo` preserved `OriginalPath`, so `ToString()` was correct.

**3. `InternetShortcutFile.IconFile` truncates paths containing `#`** — `00d05cc68`
`Uri.TryCreate(iconFile, Absolute) && IsFile ? LocalPath : iconFile` — a path like `C:\Icons\C#\app.ico` parses as a URI with `C#\app.ico`... as a fragment, and `LocalPath` returns `C:\Icons\C`. Literal `%xx` sequences in a path also get percent-decoded.

**4. Unreachable `throw;` left in** — `8982988f4`
`WindowUtilities.cs:141-142` and #L167-L168 both have `return null;` immediately followed by `throw;`. Dead code, still present at HEAD.

**5. `d2141ba3f` replaced PATH lookup with a hard-coded System32** — the removed code in `Uninstall-ADTApplication.ps1` used `Get-Command` (full PATH search). The replacement in `InstalledApplication` only tries `%SystemRoot%\System32`. An uninstall string naming a program that lives elsewhere on PATH now yields a non-existent path.

**6. `Get-ADTPEFileArchitecture` reuses a 4 KB buffer across pipeline items** — `d10ec50b1` added good signature checks, but `$data` is allocated once in `begin` and `$stream.Read()`'s return is discarded, so a file smaller than 4096 bytes leaves stale bytes from the previous file. The new `peHeaderOffset` bounds check can still point into that stale data.

## Breaking changes (public API / behaviour)

| Commit | Change |
|---|---|
| `e78f0c8dc` | `EnvironmentUtilities.GetEnvironmentVariables()` return type `IDictionary` → `IReadOnlyDictionary<string,string?>`; whitespace values become `null`; uses `.Add()` so a duplicate-cased name throws |
| `c9dee97bb` | `[ValidateNotNullOrWhiteSpace()]` now validates **dictionary values**. Any hashtable parameter with a null/empty value now fails. `8aa0841b8` had to fix `New-ADTDialogOptionsObject` because of exactly this — worth auditing every remaining use |
| `3d8ceb840` | `TimeSpanTransformationAttribute`: `"90"` was 90 **days**, now 90 **seconds**. Also dropped `AllowThousands`; `"NaN"`/`"Infinity"` now reach `TimeSpan.FromSeconds` and throw |
| `64d39f179` | `EnvironmentTable` `record class` → `class` — loses value equality, `with`, and record `ToString()`. It's public (`$adtEnv`) |
| `7bde61dc4` | **Removed** `InternetShortcutFile/Info.WorkingDirectory` and `.ShowCommand`. `Set-ADTShortcut -WorkingDirectory/-WindowStyle/-Clear WorkingDirectory` on a `.url` is now a silent no-op, and those values are no longer returned by `Get-ADTShortcut`. They're still in the `-Clear` `ValidateSet` |
| `9be647336` | `ShellLinkFile.Resolve()` `void` → `LINK_STATUS`. In PowerShell `$link.Resolve()` now emits to the pipeline. `LINK_STATUS_UNRESOLVED` is never returned |
| `25f86e673` | `RestartDialogOptions.AllowCancel` → `DialogAllowCancel` (module updated consistently) |
| `41690b427` | `[Flags]` removed from `DialogBoxButtons`/`DefaultButton`/`Icon` — `-band`/`HasFlag`/comma-string parsing no longer work |
| `33502d8ed` | `MsiUtilities.GetExceptionForMsiExitCode(int)` → `(uint)` |
| `fbc9155f8` | `DeferHistory` positional record gains a 4th parameter |
| `d407ee59b` | `AppxManifestInfo.ResourceId` `string` → `string?` |
| `140572e67`, `448f658b9`, `f81e79d5e` | ~20 public `FileInfo`/`DirectoryInfo` fields/auto-properties became computed properties backed by a `.FullName` string. Paths are now normalised at construction, and a **new object is allocated on every read** |
| `919d3cc4e` | `RebootInfo.HasPendingReboot()` now includes `IsIntuneClientRebootPending` — more deployments will be blocked |
| `9dc01629c` | Wildcard registry paths in `Remove-ADTRegistryKey` now actually delete (previously the guard always short-circuited) |
| `fcd07db51` | `Test-ADTServiceExists -Name` now accepts wildcards without `-UseCIM` |
| `9075242ca` | `DoubleSpaceRegex.Replace(title, "")` → `" "` — correct, but `InstallName` and therefore log file names change for any app with double spaces |
| `e88495d7e` | `LogFileNameRegex` is now case-insensitive — more files matched during rotation/archival |
| `ef8b1feec` | `ValidateUniqueAttribute` applies `StringComparison` per-pair — strictly more collections now fail |

## Lower-severity observations

- **Perf**: `42afddec1` made `ArgumentList`, `StandardInput`, `ListItems`, `ContainedElementRecords`, `ParentProcessFilter` etc. copy the whole list on every property get. `ProcessLaunchInfo.cs:207` makes three copies in one expression; any indexed loop over these is now O(n²).
- **Residual `{0}` risk**: `c979c50f2` replaced `string.Format` with `.Replace("{0}", …)` to stop braces in `InstallName` being read as format items — but an `InstallName` that literally contains `{0}` will now have the discriminator substituted into it (braces are legal in file names).
- `75debc9c7` mutates the caller's `$AccessControlList` object in place via an SDDL round-trip; that can lose non-canonical ACE ordering.
- `e6f7a134f` changed `NullableExtensions.ToPointer<T>` to `in this` to stop the pointer dangling — but C# still materialises a temporary for an rvalue receiver, so the hazard isn't fully closed by the signature alone.
- `631791af4` silently changed `SafeMemoryHandle.Write(byte[], int)` from "start index in source" to "offset in destination"; only one in-repo caller (offset 0), so no live impact.
- `4d9dc2039` leaks the `WindowsInstaller.Installer` COM object if `process` throws terminating (release is only in `end`).
- `005698aca` added `-ErrorAction Ignore` broadly, which will also silently skip resolvable profiles when a DC is unreachable.
- `115360d9c` → `332eba4f6` → `fb676066b` → `7bde61dc4` → `f29914819` is add-then-remove churn on internet-shortcut icon/working-directory handling; the net result is the removal in the table above.

Genuinely good fixes worth calling out: `cce31fefd` (missing `process{}` block meant the GPO account fallback only ever examined one item), `9d525f90b` (missing `else`), `80475c0ec` (reflection on `IReadOnlyDictionary.Count` always returned 0, so those parameters always threw), `c1cc563aa` (BOM on stdin), `69780a875` (`unchecked` — the repo builds with `CheckForOverflowUnderflow=true`, so `Load64` was throwing for every real address), and `4d9dc2039` (MSI SQL injection via unescapable quotes).

mjr4077au:
Hi @DanGough,

I've gone through everything and have addressed all concerns. In addition, some other issues were identified and resolved also.

## Responses to direct remarks

> Found an issue pre-existing in RC1, found here when testing changes to SecureArgumentList handling in header functions (error only shows with below command when used with `-SecureArgumentList`):
> 
> ```powershell
> Start-ADTProcess -FilePath C:\Windows\System32\cmd.exe -ArgumentList '/c exit 1' -SecureArgumentList
> ```
> 
> ```
> Invoke-ADTFunctionErrorHandler : Parameter set cannot be resolved using the specified named parameters.
> At C:\Users\admin\Desktop\PSAppDeployToolkit\Public\Start-ADTProcess.ps1:1098 char:17
> +                 Invoke-ADTFunctionErrorHandler @iafehParams -Silent
> ```

This one comes up a lot, but `-SecureArgumentList` isn't a switch. You either use one or the other. Been that way since the 3.x days.

> Another thing (also found in AI report below) was that Uninstall-ADTApplication now only looks under System32 if the uninstaller path is not fully qualified, when before it searched the whole system path. I can put notepad as my uninstall string and add/remove programs will execute it, as did PSADT 4.1; now it complains Could not find file 'C:\WINDOWS\system32\notepad'.

This was actually deliberate. The only installers I've ever seen that do this are MSIs reaching for `msiexec.exe`. The direct pinning to the system32 directory is to stop potential execution of a process as SYSTEM via path resolution which may not be what is intended to be launched. We do the same pinning when loading libraries (DLL files) in also, such as: https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/blob/c92d12846f6f68b781a20c911bb28e23b8a7c119/src/PSADT/PSADT.Interop/NativeMethods.cs#L1338

If you know of any other installation type that does this and may be expecting to fall back to path-based resolution, I'm happy to revert it but as it stands, it's a patched security problem that I'd prefer to leave unless demonstrated as limiting in a real-life scenario.

## Likely bugs (highest confidence)

> -Append/-Remove read from the wrong scope — data-loss risk on machine PATH — ceefbad6e
EnvironmentUtilities.cs:186 and #L211 compute the append/remove result from GetEnvironmentVariable(variable) — the process environment — then write it to the target registry hive. So Set-ADTEnvironmentVariable -Variable PATH -Value X -Append -Target Machine writes the process's fully-expanded PATH (which includes the user hive and anything the session added) into HKLM. The read was already target-less before this commit, but the commit's stated intent ("append/remove now apply to a process variable the same way as user or machine") makes the mismatch worse, not better. Should be GetEnvironmentVariable(variable, target).

Confirmed and resolved.

> ShellLinkInfo.IconLocation mangles environment-variable icon paths — 448f658b9
IconLocationValue = shellLink.IconLocation?.FullName. IShellLinkW::GetIconLocation returns the raw string, which is very commonly %SystemRoot%\System32\imageres.dll. Taking .FullName prefixes the current working directory, so Get-ADTShortcut now reports D:\SomeCwd\%SystemRoot%\System32\imageres.dll. Previously the FileInfo preserved OriginalPath, so ToString() was correct.

Confirmed and resolved.

> InternetShortcutFile.IconFile truncates paths containing # — 00d05cc68
Uri.TryCreate(iconFile, Absolute) && IsFile ? LocalPath : iconFile — a path like C:\Icons\C#\app.ico parses as a URI with C#\app.ico... as a fragment, and LocalPath returns C:\Icons\C. Literal %xx sequences in a path also get percent-decoded.

Could not corroborate, however it exposed another issue that has been addressed.

> Unreachable throw; left in — 8982988f4
WindowUtilities.cs:141-142 and #L167-L168 both have return null; immediately followed by throw;. Dead code, still present at HEAD.

Deliberate for CA1031 suppression.

> d2141ba3f replaced PATH lookup with a hard-coded System32 — the removed code in Uninstall-ADTApplication.ps1 used Get-Command (full PATH search). The replacement in InstalledApplication only tries %SystemRoot%\System32. An uninstall string naming a program that lives elsewhere on PATH now yields a non-existent path.

Discussed above.

> Get-ADTPEFileArchitecture reuses a 4 KB buffer across pipeline items — d10ec50b1 added good signature checks, but $data is allocated once in begin and $stream.Read()'s return is discarded, so a file smaller than 4096 bytes leaves stale bytes from the previous file. The new peHeaderOffset bounds check can still point into that stale data.

Confirmed and resolved.

## Breaking changes (public API / behaviour)

Everything here is deliberate and intentional.

## Lower-severity observations

> Perf: 42afddec1 made ArgumentList, StandardInput, ListItems, ContainedElementRecords, ParentProcessFilter etc. copy the whole list on every property get. ProcessLaunchInfo.cs:207 makes three copies in one expression; any indexed loop over these is now O(n²).

Genuine. Have refactored this setup to eliminate these copies.

> Residual {0} risk: c979c50f2 replaced string.Format with .Replace("{0}", …) to stop braces in InstallName being read as format items — but an InstallName that literally contains {0} will now have the discriminator substituted into it (braces are legal in file names).

Genuine, using a unique sentinel as a replacement marker.

> 75debc9c7 mutates the caller's $AccessControlList object in place via an SDDL round-trip; that can lose non-canonical ACE ordering.

Difficult to repro in real-life circumstances, but measures taken all the same.

> e6f7a134f changed NullableExtensions.ToPointer<T> to in this to stop the pointer dangling — but C# still materialises a temporary for an rvalue receiver, so the hazard isn't fully closed by the signature alone.

Confirmed. Have refactored to turn scenario into compilation error.

> 631791af4 silently changed SafeMemoryHandle.Write(byte[], int) from "start index in source" to "offset in destination"; only one in-repo caller (offset 0), so no live impact.

Deliberate.

> 4d9dc2039 leaks the WindowsInstaller.Installer COM object if process throws terminating (release is only in end).

Confirmed and have refactored. This function is on the chopping block anyway, but good to have it at its best for its final hurrah.

> 005698aca added -ErrorAction Ignore broadly, which will also silently skip resolvable profiles when a DC is unreachable.

Am now only specifically catching specific exception types (I think this is what it wanted here).

## Conclusion

Let me know if there's any further concerns or whether this addresses everything.

neo42JBR:
Chiming in on uninstaller path resolution:

For reference take winget-cli behaviour. Any install (machine and user scope) will generate this uninstall string:
`winget uninstall --product-code <ID>`

Where winget will not reside in system32.

DanGough:
Thanks @neo42JBR I was not aware of this - confirmed it at least does this for portable apps:

image

I had wondered if there'd be some rare edge cases out there calling powershell/java/python/node/etc without a qualified path - but package managers are another thing that could exhibit this behaviour.

mjr4077au:
Thanks @neo42JBR and @DanGough for advisement and confirmation, respectively. I've backed that stuff out.

DanGough:
## Verified fixed

| Report | Commit | Verdict |
|---|---|---|
| `-Append`/`-Remove` wrong scope | `64ec07da8` | Fixed — appending to a User var no longer pulls in the process value (tested) |
| `ShellLinkInfo.IconLocation` mangled | `5b3847f15` | Fixed — `%SystemRoot%\System32\imageres.dll` round-trips, and setting `IconIndex` alone no longer corrupts the location |
| `InternetShortcutFile.IconFile` truncation | `3044eac00` | Fixed and better than before — `%SystemRoot%\…`, `C#`, literal `%20`, UNC and `https://` all round-trip byte-for-byte |
| PE 4 KB buffer reuse | `bd3930c32` | Fixed — a 2-byte file after a PE file is correctly rejected |
| MSI installer COM leak | `dcc2219e9` | Correct; the `begin`-defined helper still resolves `$installer` because `begin`/`process` share one scope |
| ACL mutated in place | `ccda1516a` | Correct; the SDDL carries the `P` protection flag, so inheritance state survives the copy |
| O(n²) collection copies | `41d781e23`, `68fa8470c` | Fixed. Verified PowerShell still binds `.Count`, the indexer and enumeration on the now-`internal` `EquatableList<T>`, and the DCS round-trip through the private `Add`/ctor is covered by tests |
| `{0}` substitution | `3d1895476` | Safe — the PUA sentinel survives `EnvironmentTable.cs:1234`, which is a blacklist of `Path.GetInvalidFileNameChars()` |
| `ToPointer<T>` rvalue hazard | `0a095cb9b` | `ref readonly` genuinely forces `in` at the call site — compile-time enforced as claimed |

## Found issues

**1. High — Log file naming change in `b575dfdbb`**
`DeploymentSession.cs:494` switched the username suffix condition from `!isAdmin` to `!AccountUtilities.CallerIsLocalSystem`. We previously used `isAdmin` since admins can overwrite other admin's log files, so the username suffix was only added for non-admins.  Appending the username to MSI log names when running as admin generated a complaint in #2301; this change does not fix that issue but instead applies the same issue to the general toolkit log.

**2. High — env var append/remove still flattens and downgrades `REG_EXPAND_SZ`.** `EnvironmentUtilities.cs:209` now reads the right *scope*, but `Environment.GetEnvironmentVariable(name, User|Machine)` expands `REG_EXPAND_SZ`. Verified against a User variable:

```
before: %SystemRoot%\A   (ExpandString)
after append 'C:\B': C:\WINDOWS\A;C:\B   (String)
```

On the machine `PATH` that bakes in `system32` and drops the value kind, since `expandable` defaults to `$false`. Read the hive with `RegistryValueOptions.DoNotExpandEnvironmentNames` and preserve the existing `RegistryValueKind` on append/remove.

**3. High — the new PATH lookup can yield an empty `FilePath`.** `Uninstall-ADTApplication.ps1:354` calls `Get-Command` with no `-CommandType`. Verified on Windows PowerShell 5.1: `Get-Command -Name 'sc'` returns the **alias** for `Set-Content`, whose `.Source` is `''` — so `sc delete MyService` style uninstall strings set `FilePath` to an empty string. Also unescaped wildcards (`[`/`]` in a path) and possible multi-result arrays. Wants `-CommandType Application`, `[WildcardPattern]::Escape()`, `Select-Object -First 1`, and a non-empty `.Source` guard.

**4. Medium — Public contract breaks**

- `InstalledApplication.UninstallStringFilePath` / `QuietUninstallStringFilePath`: `FileInfo?` → `string?` (`7d1fa278d`)
- `ShellLinkFile.IconLocation` and `ShellLinkInfo.IconLocation`: `FileInfo?` → `string?` (`5b3847f15`)

This was necessary, but will need to be documented in case anybody was making use of `App.UninstallStringFilePath.FullName` etc.

- `IniUtilities.GetSection`: `OrderedDictionary?` → `OrderedDictionary` (`35c07bf24`)

Get-ADTIniSection on an empty section now returns null instead of an empty dictionary. Unlikely to cause any issues out there but was this change necessary?

**5. Medium — the narrowed catch in `Get-ADTUserProfiles.ps1:193` misses the case it targets.** `ConvertTo-ADTNTAccountOrSID`'s SID fallback calls `GroupPolicyAccountInfo::Get()`, which throws when not elevated — verified: `SecurityException: Requested registry access is not allowed`. That *replaces* the `IdentityNotMappedException` before the bare `throw` is reached, so `catch [IdentityNotMappedException]` doesn't match and `Get-ADTUserProfiles` now hard-fails where it previously skipped. (PowerShell's catch matching does walk nested inner exceptions, so the type itself is right.) The fix belongs in `ConvertTo-ADTNTAccountOrSID.ps1:148` — guard the GP-cache lookup so its own failure can't mask the original error.

**6. Low — `IconFile` getter can now throw.** The remark says a value that is genuinely a URL "is returned as it came", but `UrlIs(…, URLIS_FILEURL)` only tests the scheme; a `` URL that `PathCreateFromUrl` rejects, or a result over `MAX_PATH`, throws out of a property getter because NativeMethods.PathCreateFromUrl throws on non-`S_OK`.

**7. Low — `FileHandleManager` drops handles more readily.** `FileHandleManager.cs:263` now also demands `PROCESS_QUERY_LIMITED_INFORMATION`, and `return`s on any name-resolution failure — so a process that grants `DUP_HANDLE` but not QLI loses *all* its handles from the result. A placeholder name would be less lossy than dropping the row.

**8. Low — `5c25b6805` is an unresolved experiment.** It reverts `Microsoft.NET.ILLink.Tasks` to `8.0.30` in the TrimHarness lock file, but that package is implicit from the SDK and pinned nowhere, so the lock file now tracks whatever the runner has and will trip `RestoreLockedMode` on any machine with a newer 8.0.x runtime pack.

**9. Nit — Stale comment in `ProcessUtilities.cs:472`**
`bbf5046bd` removed the `processIdInfoPtr.Clear()` (correctly — it's replaced by `processIdInfo = new() { ProcessId = … }`), but the line above still reads *"We must clear the buffer ourselves as stackalloc buffers are undefined."* The comment now describes something the code no longer does.

**10. Nit — `HelpConsoleOptions.ModuleHelpMap` is the odd one out**
`41d781e23` stopped every other collection property copying per read, but `HelpConsoleOptions.cs:57` went the other way in `8ed5ad589` — it now does an outer `ToDictionary` plus an inner `ToDictionary` on *every* get. It's forced, because `EquatableDictionary<string, EquatableDictionary<string, string>>` isn't assignable to `IReadOnlyDictionary<string, IReadOnlyDictionary<string, string>>` (class generics aren't covariant). Read once per dialog, so it costs nothing in practice — but if the intent is consistency, typing the field as `EquatableDictionary<string, IReadOnlyDictionary<string, string>>` would let it hand the value out directly like its siblings.

**11. Nit — `$null = $null` in `Get-ADTUserProfiles.ps1:200`**
The catch body only needs to produce nothing so the assignment lands `$null` in `$ntAccount`. An empty block does that on its own; `$null = $null` reads like a mistake someone will later "fix".

mjr4077au:
> **1. High — Log file naming change in `b575dfdbb`**. `DeploymentSession.cs:494` switched the username suffix condition from `!isAdmin` to `!AccountUtilities.CallerIsLocalSystem`. We previously used `isAdmin` since admins can overwrite other admin's log files, so the username suffix was only added for non-admins.  Appending the username to MSI log names when running as admin generated a complaint in #2301; this change does not fix that issue but instead applies the same issue to the general toolkit log.

This has been reverted. This was intended for something in relation for #2301.

> **2. High — env var append/remove still flattens and downgrades `REG_EXPAND_SZ`.** `EnvironmentUtilities.cs:209` now reads the right *scope*, but `Environment.GetEnvironmentVariable(name, User|Machine)` expands `REG_EXPAND_SZ`. Verified against a User variable:

Remediated, with extra tests included.

> **3. High — the new PATH lookup can yield an empty `FilePath`.** `Uninstall-ADTApplication.ps1:354` calls `Get-Command` with no `-CommandType`. Verified on Windows PowerShell 5.1: `Get-Command -Name 'sc'` returns the **alias** for `Set-Content`, whose `.Source` is `''` — so `sc delete MyService` style uninstall strings set `FilePath` to an empty string. Also unescaped wildcards (`[`/`]` in a path) and possible multi-result arrays. Wants `-CommandType Application`, `[WildcardPattern]::Escape()`, `Select-Object -First 1`, and a non-empty `.Source` guard.

Remediated. This bug was always there as the initial revert just put it back. Now it's fixed properly.

> **4. Medium — Public contract breaks**. `IniUtilities.GetSection`: `OrderedDictionary?` → `OrderedDictionary` (`35c07bf24`): Get-ADTIniSection on an empty section now returns null instead of an empty dictionary. Unlikely to cause any issues out there but was this change necessary?

This is completely false. `IniUtilities.GetSection()` has a return signature that indicated it may return null, but it never does. If what the above says is true, dropping the nullability on the return type wouldn't even compile.

> **5. Medium — the narrowed catch in `Get-ADTUserProfiles.ps1:193` misses the case it targets.** `ConvertTo-ADTNTAccountOrSID`'s SID fallback calls `GroupPolicyAccountInfo::Get()`, which throws when not elevated — verified: `SecurityException: Requested registry access is not allowed`. That *replaces* the `IdentityNotMappedException` before the bare `throw` is reached, so `catch [IdentityNotMappedException]` doesn't match and `Get-ADTUserProfiles` now hard-fails where it previously skipped. (PowerShell's catch matching does walk nested inner exceptions, so the type itself is right.) The fix belongs in `ConvertTo-ADTNTAccountOrSID.ps1:148` — guard the GP-cache lookup so its own failure can't mask the original error.

This has been addressed.

> **6. Low — `IconFile` getter can now throw.** The remark says a value that is genuinely a URL "is returned as it came", but `UrlIs(…, URLIS_FILEURL)` only tests the scheme; a `` URL that `PathCreateFromUrl` rejects, or a result over `MAX_PATH`, throws out of a property getter because NativeMethods.PathCreateFromUrl throws on non-`S_OK`.

This is OK.

> **7. Low — `FileHandleManager` drops handles more readily.** `FileHandleManager.cs:263` now also demands `PROCESS_QUERY_LIMITED_INFORMATION`, and `return`s on any name-resolution failure — so a process that grants `DUP_HANDLE` but not QLI loses *all* its handles from the result. A placeholder name would be less lossy than dropping the row.

This is also OK. Requesting `PROCESS_QUERY_LIMITED_INFORMATION` has a lower requirement threshold than requesting `DUP_HANDLE` so the stated problem won't occur.

> **8. Low — `5c25b6805` is an unresolved experiment.** It reverts `Microsoft.NET.ILLink.Tasks` to `8.0.30` in the TrimHarness lock file, but that package is implicit from the SDK and pinned nowhere, so the lock file now tracks whatever the runner has and will trip `RestoreLockedMode` on any machine with a newer 8.0.x runtime pack.

Remediated. `Microsoft.NET.ILLink.Tasks` is now under CPM.

> **9. Nit — Stale comment in `ProcessUtilities.cs:472`** `bbf5046bd` removed the `processIdInfoPtr.Clear()` (correctly — it's replaced by `processIdInfo = new() { ProcessId = … }`), but the line above still reads *"We must clear the buffer ourselves as stackalloc buffers are undefined."* The comment now describes something the code no longer does.

Addressed.

> **10. Nit — `HelpConsoleOptions.ModuleHelpMap` is the odd one out** `41d781e23` stopped every other collection property copying per read, but `HelpConsoleOptions.cs:57` went the other way in `8ed5ad589` — it now does an outer `ToDictionary` plus an inner `ToDictionary` on *every* get. It's forced, because `EquatableDictionary<string, EquatableDictionary<string, string>>` isn't assignable to `IReadOnlyDictionary<string, IReadOnlyDictionary<string, string>>` (class generics aren't covariant). Read once per dialog, so it costs nothing in practice — but if the intent is consistency, typing the field as `EquatableDictionary<string, IReadOnlyDictionary<string, string>>` would let it hand the value out directly like its siblings.

Also addressed.

> **11. Nit — `$null = $null` in `Get-ADTUserProfiles.ps1:200`** The catch body only needs to produce nothing so the assignment lands `$null` in `$ntAccount`. An empty block does that on its own; `$null = $null` reads like a mistake someone will later "fix".

Awkward looking, but deliberate. We can't have an empty catch block as PSScriptAnalyzer complains, but `$null = $null` is meant to corerce the statement to return an AutomationNull, which can affect pipelining.

> Found an issue pre-existing in RC1, found here when testing changes to SecureArgumentList handling in header functions (error only shows with below command when used with -SecureArgumentList).

Also addressed.

DanGough:
## Findings

**1. `Convert-ADTNTAccountToSID` has the identical bug that `7c3eb3008` fixed in the other direction.** `ConvertTo-ADTNTAccountOrSID.ps1:183` still calls `GroupPolicyAccountInfo::Get()` inside the `if` condition, so its own failure replaces the original error. Verified side by side, non-elevated:

```
SID -> NTAccount (fixed):        MATCHED IdentityNotMappedException
NTAccount -> SID (unchanged):    FELL THROUGH: System.Security.SecurityException
```

**2. `-Append -Expandable` onto an existing `REG_SZ` stores a `%VAR%` that will never expand.** `EnvironmentUtilities.cs:384` always prefers the existing kind, so the caller's explicit `-Expandable` is discarded:

```
String + '-Append -Expandable' -> Kind=String Raw=[C:\A;%SystemRoot%\B]
```

Only the *demotion* (`ExpandString` → `String`) is destructive; the promotion is safe and is what the caller asked for. Preferring `ExpandString` when either side wants it would close this.

**3. Dead known type.** `DataSerialization.cs:419` still registers `EquatableDictionary<string, EquatableDictionary<string, string>>`, a shape that no longer exists now that `ModuleHelpMapValue` holds the inner dictionaries as the interface. Harmless — the round-trip works via the new `[KnownType]` on the options record — but it's stale.

**4. Nits in the new code**
- Trailing `;` after `return` in both `Test-ADTCallerIsSystem.ps1:12` and `Get-ADTCallerUserName.ps1:12` — C# habit, not the file's style.
- `Invoke-ADTDotNetCompilation` logs *"Locating Git on this system to determine whether debug DLLs require compilation"* unconditionally, including when running as SYSTEM where git is never consulted.

mjr4077au:
@DanGough all remediated.

DanGough (approved):

PR #2345, opened 2026-09-23
Add retry support to our two PSScriptAnalyzer functions.

mjr4077au:
# Pull Request

## Description

We often have build failures due to PSScriptAnalyzer having issues. It's widely known and documented on https://github.com/PowerShell/PSScriptAnalyzer/issues/1867. The only option we have is to retry, which usually always works.

This adds retrying directly into our code, with the idea that the build system retries once and just gets on with it. Not a great solution but all we can do.

## Type of change

## Checklist


DanGough (approved):
Let's see if this resolves things, we can consider a retry loop next if it might need to retry twice or more.

